PR #2, opened 2022-04-21
adding gcp enclave tool page

tqinli:
This is a static html page without any backend interaction, can be used to generate GCP EnclaveID and gcloud command line for GCP Enclave provisioning.


atarassov-ttd (approved):
LGTM

PR #7, opened 2022-07-19
Admin Logging Implementation for AdminKeyService

pandabear15:
Added implementations of various interfaces, and added logging capabilities to AdminKeyService.java. Tests are written, but currently would break the current workflow as the QLDB is not currently mocked and therefore would not work without a QLDB key. Tests will be pushed as soon as tests that don't break the workflow are made. The logging functionality also will not work for a similar reason until a key is made.

### DIFF
--- a/conf/default-config.json
+++ b/conf/default-config.json
@@ -1,4 +1,5 @@
 {
+  "enable_admin_logging": false,
   "qldb_ledger_name": "uid2-audit-logs",
   "qldb_table_name": "logs"
 }
--- a/conf/default-config.json
+++ b/conf/default-config.json
@@ -1,2 +1,4 @@
 {
+  "qldb_ledger_name": "uid2-audit-logs",
+  "qldb_table_name": "logs"
 }
--- a/conf/local-config.json
+++ b/conf/local-config.json
@@ -1,2 +1,4 @@
 {
+  "qldb_ledger_name": "audit-logs",
+  "qldb_table_name": "logs"
 }
--- a/conf/local-config.json
+++ b/conf/local-config.json
@@ -1,5 +1,5 @@
 {
-  "enable_admin_logging": true,
+  "enable_admin_logging": false,
   "qldb_ledger_name": "audit-logs",
   "qldb_table_name": "logs"
 }
--- a/conf/local-config.json
+++ b/conf/local-config.json
@@ -1,4 +1,5 @@
 {
+  "enable_admin_logging": true,
   "qldb_ledger_name": "audit-logs",
   "qldb_table_name": "logs"
 }
--- a/pom.xml
+++ b/pom.xml
@@ -90,11 +90,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
-            <version>1.7.26</version>
-        </dependency>
-        <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-core</artifactId>
             <version>1.2.9</version>
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,31 @@
             <version>${vertx.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.26</version>
+        </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-core</artifactId>
+            <version>1.2.9</version>
+        </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>1.2.9</version>
+        </dependency>
+        <dependency>
+            <groupId>com.github.loki4j</groupId>
+            <artifactId>loki-logback-appender</artifactId>
+            <version>1.3.2</version>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.qldb</groupId>
+            <artifactId>amazon-qldb-driver-java</artifactId>
+            <version>2.3.1</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/pom.xml
+++ b/pom.xml
@@ -90,21 +90,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>ch.qos.logback</groupId>
-            <artifactId>logback-core</artifactId>
-            <version>1.2.9</version>
-        </dependency>
-        <dependency>
-            <groupId>ch.qos.logback</groupId>
-            <artifactId>logback-classic</artifactId>
-            <version>1.2.9</version>
-        </dependency>
-        <dependency>
-            <groupId>com.github.loki4j</groupId>
-            <artifactId>loki-logback-appender</artifactId>
-            <version>1.3.2</version>
-        </dependency>
-        <dependency>
             <groupId>software.amazon.qldb</groupId>
             <artifactId>amazon-qldb-driver-java</artifactId>
             <version>2.3.1</version>
--- a/src/main/java/com/uid2/admin/Constants.java
+++ b/src/main/java/com/uid2/admin/Constants.java
@@ -1,7 +1,5 @@
 package com.uid2.admin;
 
 public class Constants {
-    public static final String QLDB_LEDGER_NAME = "audit-logs";
-    public static final String QLDB_TABLE_NAME = "logs";
     public static final String DEFAULT_ITEM_KEY = "$all_items&"; //needs to be unlikely to be a name of a current or future user
 }
--- a/src/main/java/com/uid2/admin/Constants.java
+++ b/src/main/java/com/uid2/admin/Constants.java
@@ -3,4 +3,5 @@ package com.uid2.admin;
 public class Constants {
     public static final String QLDB_LEDGER_NAME = "audit-logs";
     public static final String QLDB_TABLE_NAME = "logs";
+    public static final String DEFAULT_ITEM_KEY = "$all_items&"; //needs to be unlikely to be a name of a current or future user
 }
--- a/src/main/java/com/uid2/admin/Constants.java
+++ b/src/main/java/com/uid2/admin/Constants.java
@@ -1,0 +1,6 @@
+package com.uid2.admin;
+
+public class Constants {
+    public static final String QLDB_LEDGER_NAME = "audit-logs";
+    public static final String QLDB_TABLE_NAME = "logs";
+}
--- a/src/main/java/com/uid2/admin/audit/AuditFactory.java
+++ b/src/main/java/com/uid2/admin/audit/AuditFactory.java
@@ -1,5 +1,9 @@
 package com.uid2.admin.audit;
 
+import io.vertx.core.json.JsonObject;
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
+
 /**
  * AuditFactory controls the instantiation/creation of AuditMiddleware objects.
  * Depending on the needs of the specific implementation, the AuditFactory
@@ -8,9 +12,9 @@ package com.uid2.admin.audit;
  * exhibit some other behavior.
  */
 public class AuditFactory {
-    private static final AuditWriter auditWriter = new QLDBAuditWriter();
-
-    public static final AuditMiddleware auditMiddleware = new QLDBAuditMiddleware(auditWriter);
+    public static AuditWriter auditWriter;
+    private static AuditMiddleware auditMiddleware;
+    private static final Logger LOGGER = LoggerFactory.getLogger(AuditFactory.class);
 
     /**
      * Returns an AuditMiddleware object for the designated class to use.
@@ -19,6 +23,15 @@ public class AuditFactory {
      * @return the designated AuditMiddleware object for the passed class.
      */
     public static AuditMiddleware getAuditMiddleware(Class<?> clazz){
+        if(auditMiddleware == null){
+            LOGGER.fatal("setConfig not called before getAuditMiddleware in class AuditFactory");
+            throw new IllegalStateException("setConfig not called before getAuditMiddleware in class AuditFactory");
+        }
         return auditMiddleware;
+    }
+
+    public static void setConfig(JsonObject config){
+        auditWriter = new QLDBAuditWriter(config);
+        auditMiddleware = new QLDBAuditMiddleware(auditWriter);
     }
 }
--- a/src/main/java/com/uid2/admin/audit/AuditFactory.java
+++ b/src/main/java/com/uid2/admin/audit/AuditFactory.java
@@ -1,0 +1,24 @@
+package com.uid2.admin.audit;
+
+/**
+ * AuditFactory controls the instantiation/creation of AuditMiddleware objects.
+ * Depending on the needs of the specific implementation, the AuditFactory
+ * can be implemented to always return the same AuditMiddleware object, create a new
+ * AuditMiddleware object for every class that calls getAuditMiddleware(Class), or
+ * exhibit some other behavior.
+ */
+public class AuditFactory {
+
+    public static final AuditMiddleware auditMiddleware = new QLDBAuditMiddleware();
+
+    /**
+     * Returns an AuditMiddleware object for the designated class to use.
+     *
+     * @param clazz the class that requires an AuditMiddleware object.
+     * @return the designated AuditMiddleware object for the passed class.
+     */
+    public static AuditMiddleware getAuditMiddleware(Class<?> clazz){
+        return auditMiddleware;
+    }
+
+}
--- a/src/main/java/com/uid2/admin/audit/AuditFactory.java
+++ b/src/main/java/com/uid2/admin/audit/AuditFactory.java
@@ -8,8 +8,9 @@ package com.uid2.admin.audit;
  * exhibit some other behavior.
  */
 public class AuditFactory {
+    private static final AuditWriter auditWriter = new QLDBAuditWriter();
 
-    public static final AuditMiddleware auditMiddleware = new QLDBAuditMiddleware();
+    public static final AuditMiddleware auditMiddleware = new QLDBAuditMiddleware(auditWriter);
 
     /**
      * Returns an AuditMiddleware object for the designated class to use.
@@ -20,5 +21,4 @@ public class AuditFactory {
     public static AuditMiddleware getAuditMiddleware(Class<?> clazz){
         return auditMiddleware;
     }
-
 }
--- a/src/main/java/com/uid2/admin/audit/AuditFactory.java
+++ b/src/main/java/com/uid2/admin/audit/AuditFactory.java
@@ -4,6 +4,9 @@ import io.vertx.core.json.JsonObject;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
 
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * AuditFactory controls the instantiation/creation of AuditMiddleware objects.
  * Depending on the needs of the specific implementation, the AuditFactory
@@ -12,26 +15,19 @@ import io.vertx.core.logging.LoggerFactory;
  * exhibit some other behavior.
  */
 public class AuditFactory {
-    public static AuditWriter auditWriter;
-    private static AuditMiddleware auditMiddleware;
+    private static final Map<JsonObject, AuditMiddleware> middlewareMap = new HashMap<>();
     private static final Logger LOGGER = LoggerFactory.getLogger(AuditFactory.class);
 
     /**
      * Returns an AuditMiddleware object for the designated class to use.
      *
-     * @param clazz the class that requires an AuditMiddleware object.
      * @return the designated AuditMiddleware object for the passed class.
      */
-    public static AuditMiddleware getAuditMiddleware(Class<?> clazz){
-        if(auditMiddleware == null){
-            LOGGER.fatal("setConfig not called before getAuditMiddleware in class AuditFactory");
-            throw new IllegalStateException("setConfig not called before getAuditMiddleware in class AuditFactory");
+    public static AuditMiddleware getAuditMiddleware(JsonObject config){
+        if(!middlewareMap.containsKey(config)){
+            middlewareMap.put(config, new AuditMiddlewareImpl(new QLDBAuditWriter(config), config));
         }
-        return auditMiddleware;
+        return middlewareMap.get(config);
     }
 
-    public static void setConfig(JsonObject config){
-        auditWriter = new QLDBAuditWriter(config);
-        auditMiddleware = new QLDBAuditMiddleware(auditWriter);
-    }
 }
--- a/src/main/java/com/uid2/admin/audit/AuditHandler.java
+++ b/src/main/java/com/uid2/admin/audit/AuditHandler.java
@@ -1,6 +1,8 @@
 package com.uid2.admin.audit;
 
+import java.util.List;
+
 @FunctionalInterface
 public interface AuditHandler<E> {
-    OperationModel handle(E var);
+    List<OperationModel> handle(E var);
 }
--- a/src/main/java/com/uid2/admin/audit/AuditHandler.java
+++ b/src/main/java/com/uid2/admin/audit/AuditHandler.java
@@ -1,8 +1,0 @@
-package com.uid2.admin.audit;
-
-import java.util.List;
-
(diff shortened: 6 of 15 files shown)


src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,141 @@
...
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+public class QLDBAuditMiddleware implements AuditMiddleware{
+
+    public static final String LEDGER_NAME = "audit-logs";
+    public static final String TABLE_NAME = "logs";
```
cody-constine-ttd, line 25: Should these too be in the const file?

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,141 @@
...
+                    }
+                }
+                if(iterator.hasNext()){
+                    query.append(", ");
+                }
+            }
+            txn.execute(query.toString(), sanitizedInputs);
+        });
+
+        // old code; performs two queries which the QLDB treats as two history entries
```
cody-constine-ttd, line 76: Removed defuncted code please

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,134 @@
...
+import software.amazon.awssdk.services.qldbsession.QldbSessionClient;
+import software.amazon.qldb.QldbDriver;
+import software.amazon.qldb.RetryPolicy;
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+public class QLDBAuditMiddleware implements AuditMiddleware{
```
atarassov-ttd, line 23: From design perspective, the logic of middleware which collects audit information should be separate from logic of storing the audit record(s) to the destination log/db.

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public static final String LEDGER_NAME = "audit-logs";
```
haitaottd, line 25: can you call it uid2-audit-logs

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,134 @@
...
+    public static final String TABLE_NAME = "logs";
+    private static final IonSystem ionSys = IonSystemBuilder.standard().build();
+    private static final QldbDriver qldbDriver = QldbDriver.builder()
```
haitaottd, line 28: don't you need username/password to access QLDB?

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Override
+    public void writeLog(AuditModel model) {
+        if(model == null){
+            return;
+        }
+        // write to QLDB; update this to UPDATE in one command instead of two
+        qldbDriver.execute(txn -> {
+            List<IonValue> sanitizedInputs = new ArrayList<>();
+            StringBuilder query = new StringBuilder("UPDATE " + TABLE_NAME + " AS t SET ");
```
haitaottd, line 53: Don't see a where clause. How does QLDB know which row to update?

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,134 @@
...
+        return auditHandler::handle;
+    }
+
+    @Override
+    public void writeLog(AuditModel model) {
+        if(model == null){
+            return;
+        }
+        // write to QLDB; update this to UPDATE in one command instead of two
+        qldbDriver.execute(txn -> {
```
haitaottd, line 51: can this throw exception

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Override
+    public Handler<RoutingContext> handle(AuditHandler<RoutingContext> handler) {
+        InnerAuditHandler auditHandler = new InnerAuditHandler(handler, this);
+        return auditHandler::handle;
+    }
+
+    @Override
+    public void writeLog(AuditModel model) {
+        if(model == null){
```
haitaottd, line 47: when would this be null

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,134 @@
...
+            if(rc.statusCode() < 300){
+                AuditModel auditModel = new QLDBAuditModel(model.itemType, model.itemKey, model.actionTaken,
+                        getIPAddress(rc),
+                        ((IAuthorizable)rc.data().get("api-client")).getContact(),
+                        System.getenv("HOSTNAME"), Instant.now().getEpochSecond(), model.itemHash, model.summary);
+                auditWriter.writeLog(auditModel);
+            }
+        }
+
+        private static String extractBearerToken(String headerValue) {
```
haitaottd, line 109: this is not used

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,134 @@
...
+                } else {
+                    String givenPrefix = v.substring(0, "bearer ".length());
+                    return !"bearer ".equalsIgnoreCase(givenPrefix) ? null : v.substring("bearer ".length());
+                }
+            }
+        }
+
+        private static String getIPAddress(RoutingContext rc){
+            List<String> listIP = rc.request().headers().getAll("X-Forwarded-For");
+            System.out.println();
```
haitaottd, line 125: remove?

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,134 @@
...
+            if(listIP.isEmpty()){
```
haitaottd, line 126: check null?

src/main/java/com/uid2/admin/vertx/service/AdminKeyService.java
```diff
@@ -73,122 +77,127 @@ public AdminKeyService(JsonObject config,
...
+        }), Role.ADMINISTRATOR));
     }
 
-    private void handleAdminMetadata(RoutingContext rc) {
+    private OperationModel handleAdminMetadata(RoutingContext rc) {
         try {
             rc.response()
                     .putHeader(HttpHeaders.CONTENT_TYPE, "application/json")
                     .end(adminUserProvider.getMetadata().encode());
+            return new OperationModel(Type.ADMIN, null, Actions.LIST,
```
haitaottd, line 136: A hardcoded key seems better. On the other hand, I don't think getting metadata is worth auditing. Same with admin list below

src/main/java/com/uid2/admin/vertx/service/AdminKeyService.java
```diff
@@ -73,122 +77,127 @@ public AdminKeyService(JsonObject config,
...
                 ResponseUtil.error(rc, 404, "admin not found");
-                return;
+                return null;
             }
 
             rc.response()
                     .putHeader(HttpHeaders.CONTENT_TYPE, "application/json")
-                    .end(jsonWriter.writeValueAsString(existingAdmin.get()));
+                    .end(jsonWriter.writeValueAsString(a));
+            return new OperationModel(Type.ADMIN, a.getName(), Actions.GET,
```
haitaottd, line 175: should key be email

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,134 @@
...
+        private final AuditHandler<RoutingContext> innerHandler;
+        private final AuditMiddleware auditWriter;
+
+        private InnerAuditHandler(AuditHandler<RoutingContext> handler, AuditMiddleware auditWriter){
+            this.innerHandler = handler;
+            this.auditWriter = auditWriter;
+        }
+
+        public void handle(RoutingContext rc){
+            OperationModel model = innerHandler.handle(rc);
```
haitaottd, line 99: check null

src/main/java/com/uid2/admin/vertx/service/AdminKeyService.java
```diff
@@ -211,112 +220,106 @@ private void handleAdminAdd(RoutingContext rc) {
...
             // delete admin from the array
-            AdminUser a = existingAdmin.get();
             admins.remove(a);
 
             // upload to storage
             storageManager.uploadAdminUsers(adminUserProvider, admins);
 
             // respond with admin deleted
             rc.response().end(jsonWriter.writeValueAsString(a));
+            return new OperationModel(Type.ADMIN, a.getName(), Actions.DELETE,
```
haitaottd, line 254: Use email as key

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
pandabear15: Reads username/pw from file on the computer/server

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,134 @@
...
+            .build();
+    private static final Logger logger = LoggerFactory.getLogger(QLDBAuditMiddleware.class);
+
+    protected QLDBAuditMiddleware(){
+
+    }
+
+    @Override
+    public Handler<RoutingContext> handle(AuditHandler<RoutingContext> handler) {
+        InnerAuditHandler auditHandler = new InnerAuditHandler(handler, this);
```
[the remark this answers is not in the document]
pandabear15: Working on removing this

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
pandabear15: service classes return null if access forbidden or other errors occur

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
pandabear15: If credentials fail; I'll put this in a try/catch

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
pandabear15: Working on fixing db design

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
pandabear15: Will do

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
pandabear15: Thank you, check added

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
pandabear15: Relic from older code

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
pandabear15: done

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
pandabear15: more relics, thanks

src/main/java/com/uid2/admin/vertx/service/AdminKeyService.java
pandabear15: According to current code, only name is guaranteed to be unique

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
pandabear15: Working on splitting logic

src/main/java/com/uid2/admin/vertx/service/AdminKeyService.java
pandabear15: I'll add a hardcoded key

src/main/java/com/uid2/admin/Constants.java
```diff
@@ -0,0 +1,7 @@
+package com.uid2.admin;
+
+public class Constants {
+    public static final String QLDB_LEDGER_NAME = "audit-logs";
```
haitaottd, line 4: Can you make these a config variable name? Value will be set as environment variable.

src/main/java/com/uid2/admin/audit/AuditFactory.java
```diff
@@ -0,0 +1,24 @@
...
+    public static final AuditMiddleware auditMiddleware = new QLDBAuditMiddleware(auditWriter);
+
+    /**
+     * Returns an AuditMiddleware object for the designated class to use.
+     *
+     * @param clazz the class that requires an AuditMiddleware object.
+     * @return the designated AuditMiddleware object for the passed class.
+     */
+    public static AuditMiddleware getAuditMiddleware(Class<?> clazz){
+        return auditMiddleware;
```
haitaottd, line 22: clazz is not used

src/main/java/com/uid2/admin/audit/AuditHandler.java
```diff
@@ -1,6 +1,8 @@
 package com.uid2.admin.audit;
 
+import java.util.List;
+
 @FunctionalInterface
 public interface AuditHandler<E> {
-    OperationModel handle(E var);
+    List<OperationModel> handle(E var);
```
haitaottd, line 7: This interface seems unnecessary. How about remove this and change AuditMiddleware interface to:
`Handler<RoutingContext> handle(Function<RoutingContext, List<OperationModel>> handler);`

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,57 @@
...
+        private final AuditHandler<RoutingContext> innerHandler;
+        private final AuditWriter auditWriter;
+        private InnerAuditHandler(AuditHandler<RoutingContext> handler, AuditWriter auditWriter) {
+            this.innerHandler = handler;
+            this.auditWriter = auditWriter;
+        }
+
+        public void writeLog(RoutingContext rc){
+            List<OperationModel> modelList = innerHandler.handle(rc);
+            if(modelList != null && rc.statusCode() < 300){
```
haitaottd, line 36: Why check status code? Shouldn't non-null operation  means something needs to be logged?

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,57 @@
...
+                for(OperationModel model : modelList) {
+                    AuditModel auditModel = new QLDBAuditModel(model.itemType, model.itemKey, model.actionTaken,
+                            getIPAddress(rc),
```
haitaottd, line 39: Get IP address before the loop so not to repeat.

src/main/java/com/uid2/admin/audit/QLDBAuditWriter.java
```diff
@@ -0,0 +1,96 @@
...
+import software.amazon.qldb.RetryPolicy;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+public class QLDBAuditWriter implements AuditWriter{
+    private static final IonSystem ionSys = IonSystemBuilder.standard().build();
+    private static final QldbDriver qldbDriver = QldbDriver.builder()
+            .ledger(Constants.QLDB_LEDGER_NAME)
```
haitaottd, line 23: Name of ledger and table should be in config instead of hard-coded

src/main/java/com/uid2/admin/audit/AuditFactory.java
pandabear15: In this implementation, yes. It's there for future-proofing; if future implementations want to have class specific loggers then this code will support that

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
pandabear15: modelList == null means there is nothing to log (this can occur if the endpoint fails due to malformed input, or auth forbids access). Status code check can be removed; this was a relic of older code

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
pandabear15: will do

src/main/java/com/uid2/admin/audit/AuditHandler.java
pandabear15: Didn't know Function was a thing; I did feel that including this interface made things awkward. Thank you for the solution

pom.xml
```diff
@@ -89,6 +89,31 @@
             <version>${vertx.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
```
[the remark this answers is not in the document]
pandabear15: Will comment out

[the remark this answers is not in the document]
haitaottd: please remove if not used

conf/default-config.json
```diff
@@ -1,2 +1,4 @@
 {
+  "qldb_ledger_name": "uid2-audit-logs",
+  "qldb_table_name": "logs"
```
haitaottd, line 3: add another config to turn auditing on/off, default to off until infra is set up

src/main/java/com/uid2/admin/audit/AuditFactory.java
haitaottd: What do you mean by class specific logger? Generally we shouldn't add unused stuff unless you know you'll need it very soon and it's troublesome to add later.

src/main/java/com/uid2/admin/audit/AuditFactory.java
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public static AuditMiddleware getAuditMiddleware(Class<?> clazz){
+        if(auditMiddleware == null){
+            LOGGER.fatal("setConfig not called before getAuditMiddleware in class AuditFactory");
+            throw new IllegalStateException("setConfig not called before getAuditMiddleware in class AuditFactory");
+        }
+        return auditMiddleware;
+    }
+
+    public static void setConfig(JsonObject config){
```
haitaottd, line 33: why don't pass config in `getAuditMiddleware`

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.time.Instant;
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Function;
+
+public class QLDBAuditMiddleware implements AuditMiddleware{
+    private final AuditWriter auditWriter;
+
+    protected QLDBAuditMiddleware(AuditWriter writer){
```
haitaottd, line 17: Now middleware is agnostic to audit storage. Better just call it AuditMIddleware

src/main/java/com/uid2/admin/audit/QLDBAuditWriter.java
```diff
@@ -0,0 +1,56 @@
...
+        qldbDriver.execute(txn -> {
+            List<IonValue> sanitizedInputs = new ArrayList<>();
+            JsonObject jsonObject = model.writeToJson();
+            StringBuilder query = new StringBuilder("UPDATE " + logTable + " AS t SET data = ?");
+            sanitizedInputs.add(ionSys.newLoader().load(jsonObject.toString()).get(0));
+            query.append(" WHERE t.data.itemType = ? AND t.data.itemKey = ?");
+            sanitizedInputs.add(ionSys.newString(jsonObject.getString("itemType")));
+            sanitizedInputs.add(ionSys.newString(jsonObject.getString("itemKey")));
+
+            Result r = txn.execute(query.toString(), sanitizedInputs);
```
haitaottd, line 47: do you need to handle exception: https://docs.aws.amazon.com/qldb/latest/developerguide/driver-cookbook-java.html#:~:text=GovId%20%3D%20%3F%22%2C%20parameters)%3B%0A%7D)%3B-,Using%20the%20Jackson%20mapper,-Note

For now, you can just log error

src/main/java/com/uid2/admin/vertx/service/AdminKeyService.java
```diff
@@ -125,70 +130,67 @@ private void handleAdminMetadata(RoutingContext rc) {
                     .putHeader(HttpHeaders.CONTENT_TYPE, "application/json")
                     .end(adminUserProvider.getMetadata().encode());
         } catch (Exception e) {
+            e.printStackTrace();
```
haitaottd, line 133: Don't do this. The service could be spammed.

src/main/java/com/uid2/admin/vertx/service/AdminKeyService.java
```diff
@@ -125,70 +130,67 @@ private void handleAdminMetadata(RoutingContext rc) {
...
+            return Collections.singletonList(new OperationModel(Type.ADMIN, Constants.DEFAULT_ITEM_KEY, Actions.LIST,
+                    null, "list admins"));
         } catch (Exception e) {
             rc.fail(500, e);
+            return null;
         }
     }
-
-    private void handleAdminReveal(RoutingContext rc) {
+    // consider constructing AuditModel here and return it, therefore AuditModel can access additional context
```
haitaottd, line 156: should this comment be removed

pandabear15: yes

conf/default-config.json
pandabear15: Andrei did say he does want to have loki auditing, so I'm going to look into it. I'll remove the dependency for now though.

src/main/java/com/uid2/admin/audit/QLDBAuditWriter.java
pandabear15: good catch, thanks

pandabear15: Turns out that we do not need to handle an exception here; the exception mentioned is only possible using a different method of writing the transaction

src/main/java/com/uid2/admin/vertx/service/AdminKeyService.java
pandabear15: Relic of older debug code, thanks

src/main/java/com/uid2/admin/audit/QLDBAuditMiddleware.java
pandabear15: Good point. AuditMiddlewareImpl

pandabear15: AuditMiddleware is already an interface

src/main/java/com/uid2/admin/audit/AuditFactory.java
pandabear15: I thought about doing that but initially judged it to be an awkward solution. Thanks for letting me know this is actually ok

conf/local-config.json
```diff
@@ -1,2 +1,5 @@
 {
+  "enable_admin_logging": true,
```
haitaottd, line 2: Should this be false too because extra setup of aws credential is needed to run locally?

pom.xml
```diff
@@ -89,6 +89,26 @@
             <version>${vertx.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
haitaottd, line 92: should these be removed?

conf/local-config.json
pandabear15: Good point, I only thought about it from my machine.

pom.xml
pandabear15: Right, we'll remove these dependencies for now while I look into reworking loki

PR #9, opened 2022-08-05
Loki Auditing

pandabear15:
In addition to logging to a QLDB, logging to Loki, a highly available logging service, makes it simple to retrieve audit logs in the future, at the cost of the lack of assurance of modification, security, etc. Adding loki auditing is not intended to replace the QLDB, indeed, the QLDB is intended to be the source of truth, and Loki logs are intended to solely be an accessible way to retrieve the same logs. If auditors need the cryptographic proof that the logs are what they should be, then they should access the QLDB instead of Loki.

### DIFF
--- a/conf/logback.loki.xml
+++ b/conf/logback.loki.xml
@@ -19,10 +19,10 @@
         </http>
         <format>
             <label>
-                <pattern>app=uid2-admin,host=${HOSTNAME},portoffset=${port_offset:-0},level=%level</pattern>
+                <pattern>app=uid2-admin,host=${HOSTNAME},portoffset=${port_offset:-0}</pattern>
             </label>
             <message>
-                <pattern>l=%level h=${HOSTNAME} po=${port_offset:-0} c=%logger{20} t=%thread | %msg %ex</pattern>
+                <pattern>%d{HH:mm:ss.SSS} | %msg</pattern>
             </message>
             <sortByTime>true</sortByTime>
         </format>
@@ -35,9 +35,12 @@
         </encoder>
     </appender>
 
+    <logger name="com.uid2.admin.audit" additivity="false">
+        <appender-ref ref="AUDIT"/>
+    </logger>
+
     <root level="DEBUG">
         <appender-ref ref="LOKI"   />
-        <appender-ref ref="AUDIT"  />
         <appender-ref ref="STDOUT" />
     </root>
 </configuration>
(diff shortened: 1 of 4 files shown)


conf/logback.loki.xml
```diff
@@ -13,6 +13,20 @@
             <sortByTime>true</sortByTime>
         </format>
     </appender>
+    <appender name="AUDIT" class="com.github.loki4j.logback.Loki4jAppender">
+        <http>
+            <url>http://localhost:3200/loki/api/v1/push</url>
```
cody-constine-ttd, line 18: Can you make a .local config if you are going to commit localhost:3200

conf/logback.loki.xml
```diff
@@ -21,8 +35,12 @@
         </encoder>
     </appender>
 
+    <logger name="com.uid2.admin.audit" additivity="false">
+        <appender-ref ref="AUDIT"/>
+    </logger>
+
     <root level="DEBUG">
-        <appender-ref ref="LOKI" />
+        <appender-ref ref="LOKI"   />
```
cody-constine-ttd, line 43: *nit* extra whitespace

conf/logback.loki.xml
pandabear15: ppppffftt this is supposed to be loki:3200

conf/logback.loki.xml
pandabear15: fixed

pom.xml
```diff
@@ -89,6 +89,21 @@
             <version>${vertx.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
haitaottd, line 92: this is probably needed too:
```
<dependency>
            <groupId>org.slf4j</groupId>
            <artifactId>slf4j-api</artifactId>
            <version>1.7.26</version>
</dependency>
```

src/main/java/com/uid2/admin/audit/QLDBAuditWriter.java
```diff
@@ -17,6 +17,7 @@
 public class QLDBAuditWriter implements AuditWriter{
     private static final IonSystem ionSys = IonSystemBuilder.standard().build();
     private static final Logger logger = LoggerFactory.getLogger(QLDBAuditWriter.class);
+    private static final Logger auditLogger = LoggerFactory.getLogger("com.uid2.admin.audit");
```
haitaottd, line 20: do you need booth loggers?

pandabear15: Yes, one logger writes to the separate loki audit log, the other is for error writing / standard loki log

pom.xml
pandabear15: will add

PR #13, opened 2022-08-22
Revisited test cases, reworked QLDBInit, added methods to IService

pandabear15:
I had made so many changes to the structure of the logs that the test cases no longer passed; this has been revisited and fixed. QLDBInit is now automatically run at startup. Methods have been added to IService for the benefit of the logging system.

### DIFF
--- a/src/main/java/com/uid2/admin/Constants.java
+++ b/src/main/java/com/uid2/admin/Constants.java
@@ -1,8 +1,5 @@
 package com.uid2.admin;
 
-import com.uid2.admin.audit.Type;
-import com.uid2.admin.vertx.service.IService;
-
 public class Constants {
     public static final String DEFAULT_ITEM_KEY = "$all_items&"; //needs to be unlikely to be a name of a current or future user
 }
--- a/src/main/java/com/uid2/admin/Constants.java
+++ b/src/main/java/com/uid2/admin/Constants.java
@@ -1,5 +1,8 @@
 package com.uid2.admin;
 
+import com.uid2.admin.audit.Type;
+import com.uid2.admin.vertx.service.IService;
+
 public class Constants {
     public static final String DEFAULT_ITEM_KEY = "$all_items&"; //needs to be unlikely to be a name of a current or future user
 }
--- a/src/main/java/com/uid2/admin/Main.java
+++ b/src/main/java/com/uid2/admin/Main.java
@@ -25,7 +25,6 @@ package com.uid2.admin;
 
 import com.uid2.admin.audit.AuditFactory;
 import com.uid2.admin.audit.AuditMiddleware;
-import com.uid2.admin.audit.AuditMiddlewareImpl;
 import com.uid2.admin.audit.QLDBInit;
 import com.uid2.admin.auth.*;
 import com.uid2.admin.secret.IKeyGenerator;
--- a/src/main/java/com/uid2/admin/Main.java
+++ b/src/main/java/com/uid2/admin/Main.java
@@ -25,6 +25,8 @@ package com.uid2.admin;
 
 import com.uid2.admin.audit.AuditFactory;
 import com.uid2.admin.audit.AuditMiddleware;
+import com.uid2.admin.audit.AuditMiddlewareImpl;
+import com.uid2.admin.audit.QLDBInit;
 import com.uid2.admin.auth.*;
 import com.uid2.admin.secret.IKeyGenerator;
 import com.uid2.admin.secret.ISaltRotation;
@@ -152,6 +154,8 @@ public class Main {
 
             AdminVerticle adminVerticle = new AdminVerticle(authHandlerFactory, auth, adminUserProvider,
                     services);
+
+            QLDBInit.init(services, config);
 
             RotatingStoreVerticle rotatingAdminUserStoreVerticle = new RotatingStoreVerticle(
                     "admins", 10000, adminUserProvider);
--- a/src/main/java/com/uid2/admin/audit/QLDBAuditWriter.java
+++ b/src/main/java/com/uid2/admin/audit/QLDBAuditWriter.java
@@ -11,8 +11,7 @@ import software.amazon.qldb.QldbDriver;
 import software.amazon.qldb.Result;
 import software.amazon.qldb.RetryPolicy;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
 
 public class QLDBAuditWriter implements AuditWriter{
     private static final IonSystem ionSys = IonSystemBuilder.standard().build();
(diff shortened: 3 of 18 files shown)


src/main/java/com/uid2/admin/Constants.java
```diff
@@ -1,5 +1,8 @@
 package com.uid2.admin;
 
+import com.uid2.admin.audit.Type;
```
cody-constine-ttd, line 3: Do you need this imports?

src/main/java/com/uid2/admin/Main.java
```diff
@@ -25,6 +25,8 @@
 
 import com.uid2.admin.audit.AuditFactory;
 import com.uid2.admin.audit.AuditMiddleware;
+import com.uid2.admin.audit.AuditMiddlewareImpl;
```
cody-constine-ttd, line 28: Do you need this import?

src/main/java/com/uid2/admin/audit/QLDBAuditWriter.java
```diff
@@ -37,20 +36,35 @@ public void writeLog(AuditModel model) {
...
-                qldbDriver.execute(txn -> {
-                    List<IonValue> sanitizedInputs = new ArrayList<>();
-                    JsonObject jsonObject = model.writeToJson();
-                    StringBuilder query = new StringBuilder("UPDATE " + logTable + " AS t SET data = ?");
-                    sanitizedInputs.add(ionSys.newLoader().load(jsonObject.toString()).get(0));
-                    query.append(" WHERE t.data.itemType = ? AND t.data.itemKey = ?");
-                    sanitizedInputs.add(ionSys.newString(jsonObject.getString("itemType")));
-                    sanitizedInputs.add(ionSys.newString(jsonObject.getString("itemKey")));
+                JsonObject jsonObject = model.writeToJson();
+                if(jsonObject.getString("actionTaken").equals("CREATE")){
```
cody-constine-ttd, line 40: Why do you write this to JSON then look up the field, couldn't you just check this field from the class?

cody-constine-ttd, line 40: Also why not have one method for created table and one for write logs? I thought you only need to create in your initialization function?

src/main/java/com/uid2/admin/vertx/service/AdminKeyService.java
```diff
@@ -124,12 +124,34 @@ public void setupRoutes(Router router) {
...
+        try {
+            Collection<AdminUser> adminUsers = adminUserProvider.getAll();
+            Collection<OperationModel> newModels = new HashSet<>();
+            for (AdminUser a : adminUsers) {
+                newModels.add(new OperationModel(Type.ADMIN, a.getName(), null,
+                        DigestUtils.sha256Hex(jsonWriter.writeValueAsString(a)), null));
+            }
+            return newModels;
+        } catch (Exception e) {
+            e.printStackTrace();
```
cody-constine-ttd, line 138: If this fails will it cause admin to fail because logging isn't setup correctly?

src/main/java/com/uid2/admin/vertx/service/AdminKeyService.java
```diff
@@ -374,7 +396,7 @@ private List<OperationModel> handleAdminRoles(RoutingContext rc) {
      * @param a the AdminUser to write
      * @return a JsonObject representing a, without the key field.
      */
-    public JsonObject adminToJson(AdminUser a){
+    public JsonObject adminToJson(AdminUser a){ // should hash key
```
cody-constine-ttd, line 399: If you mean to hash the key then do it? Don't just leave a comment to do it.

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -136,6 +136,27 @@ public void setupRoutes(Router router) {
         }), Role.CLIENTKEY_ISSUER));
     }
 
+    @Override
+    public Collection<OperationModel> qldbSetup(){
+        try {
+            Collection<ClientKey> clients = clientKeyProvider.getAll();
```
cody-constine-ttd, line 142: This is duplicate code from above.

src/main/java/com/uid2/admin/vertx/service/EnclaveIdService.java
```diff
@@ -88,6 +88,27 @@ public void setupRoutes(Router router) {
         }), Role.ADMINISTRATOR));
     }
 
+    @Override
+    public Collection<OperationModel> qldbSetup(){
+        try {
+            Collection<EnclaveIdentifier> enclaves = enclaveIdProvider.getAll();
```
cody-constine-ttd, line 94: Also duplicate logic

src/main/java/com/uid2/admin/vertx/service/PartnerConfigService.java
```diff
@@ -71,6 +71,25 @@ public void setupRoutes(Router router) {
         }), Role.ADMINISTRATOR));
     }
 
+    @Override
+    public Collection<OperationModel> qldbSetup(){
+        try {
+            String config = partnerConfigProvider.getConfig();
```
cody-constine-ttd, line 77: I think you could write one method that can handle all these cases, so the logic is all in one place.

src/main/java/com/uid2/admin/audit/QLDBAuditWriter.java
pandabear15: We need to create a new entry in the event that for example a new admin is created; it must now have a corresponding QLDB entry

src/main/java/com/uid2/admin/vertx/service/AdminKeyService.java
pandabear15: I'll come around to address this case, but this falls under the category of failure cases which I'm planning to handle along with some other things in a future PR. For now I just want to see if I can get the integ environment up and running at all

src/main/java/com/uid2/admin/vertx/service/AdminKeyService.java
pandabear15: woops that slipped, thanks

src/main/java/com/uid2/admin/audit/QLDBAuditWriter.java
pandabear15: I either do this or assert that model instanceof QLDBAuditWriter. To be fair, this is probably a fair assertion as there's something wrong if a different audit model is being passed :)

cody-constine-ttd (approved):

PR #19, opened 2022-11-22
Adding code coverage

cody-constine-ttd:


pom.xml
```diff
@@ -147,6 +147,18 @@
                     </systemProperties>
                 </configuration>
             </plugin>
+            <plugin>
```
jon8787, line 150: I'm pretty keen for a parent pom :)

PR #21, opened 2022-12-15
Added site ID to operator key APIs

gmsdelmundo-ttd:


gmsdelmundo-ttd:
Setting this PR as a draft until the uid2-shared version is updated.

src/test/java/com/uid2/admin/vertx/OperatorKeyServiceTest.java
```diff
@@ -0,0 +1,98 @@
...
+            } catch (Exception ex) {
+                fail(ex);
+            }
+
+            testContext.completeNow();
+        });
+    }
+
+    @Test
+    void clientAddUnknownSiteId(Vertx vertx, VertxTestContext testContext) {
```
sunnywu, line 63: these are still tests to test adding/changing operator/operatorKey, the term "client" is a bit confusing as it seems to refer to Client Key so maybe rename it back to operatorAddUnknownSiteId instead?

gmsdelmundo-ttd: Sure, will updating this

sunnywu (approved):

src/main/java/com/uid2/admin/vertx/service/SaltService.java
```diff
@@ -61,6 +61,7 @@ private void handleSaltSnapshots(RoutingContext rc) {
                     .putHeader(HttpHeaders.CONTENT_TYPE, "application/json")
                     .end(ja.encode());
         } catch (Exception e) {
+            e.printStackTrace();
```
sunnywu, line 64: nitpick - do you still want these printStackTrace() above/below, or do `LOGGER.error(e.getMessage(), e); `instead?

sunnywu (approved):

PR #22, opened 2022-12-23
Uid2-598 differentiate public private operator

sunnywu:
Create new operator key functionalities to update OperatorKey.publicOperator field:

image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-admin</artifactId>
-    <version>1.2.1</version>
+    <version>1.1.1</version>
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
         <!-- check micrometer.version vertx-micrometer-metrics consumes before bumping up -->
         <micrometer.version>1.1.0</micrometer.version>
         <junit-jupiter.version>5.7.0</junit-jupiter.version>
-        <uid2-shared.version>1.12.0</uid2-shared.version>
+        <uid2-shared.version>1.13.0</uid2-shared.version>
         <image.version>${project.version}</image.version>
     </properties>
     <dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
         <!-- check micrometer.version vertx-micrometer-metrics consumes before bumping up -->
         <micrometer.version>1.1.0</micrometer.version>
         <junit-jupiter.version>5.7.0</junit-jupiter.version>
-        <uid2-shared.version>1.121.0-SNAPSHOT</uid2-shared.version>
+        <uid2-shared.version>1.122.0-SNAPSHOT</uid2-shared.version>
         <image.version>${project.version}</image.version>
     </properties>
     <dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-admin</artifactId>
-    <version>1.1.1</version>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -16,7 +16,7 @@
         <!-- check micrometer.version vertx-micrometer-metrics consumes before bumping up -->
         <micrometer.version>1.1.0</micrometer.version>
         <junit-jupiter.version>5.7.0</junit-jupiter.version>
-        <uid2-shared.version>1.11.0</uid2-shared.version>
+        <uid2-shared.version>1.122.0-SNAPSHOT</uid2-shared.version>
```
sunnywu, line 19: ignore will fix before this PR is merged.

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
```diff
@@ -128,6 +133,8 @@ private void handleOperatorList(RoutingContext rc) {
                 jo.put("created", o.getCreated());
                 jo.put("disabled", o.isDisabled());
                 jo.put("site_id", o.getSiteId());
+                jo.put("is_public_operator", o.isPublicOperator());
```
aulme, line 136: I don't think having both `is_public_operator` and `is_private_operator` is a good idea. They can't be both false or both true, this is just error cases that now would need to be handled.  My preference would be to have a something like `operatorType` and that could have values of `public` and `private`. This would only have one error condition (the type being unknown/null) and would allow for extensibility if we have another operator type in the future (e.g. shared). This would also remove the need for the awkward `equals("true")` below.

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
```diff
@@ -309,6 +321,45 @@ private void handleOperatorDisable(RoutingContext rc, boolean disableFlag) {
...
+            }
+
+            List<OperatorKey> operators = this.operatorKeyProvider.getAll()
+                    .stream().sorted((a, b) -> (int) (a.getCreated() - b.getCreated()))
+                    .collect(Collectors.toList());
+
+            boolean isPublicOperator = rc.queryParam("isPublicOperator").get(0) == null ? false :
+                    rc.queryParam("isPublicOperator").get(0).equals("true") ? true : false;
+
+            OperatorKey c = existingOperator.get();
```
aulme, line 348: Don't thing we should name variables with one letter outside of perhaps a one line lambda param. Not sure why it's `c` actually.

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
```diff
@@ -97,6 +96,12 @@ public void setupRoutes(Router router) {
             }
         }, Role.ADMINISTRATOR));
 
+        router.post("/api/operator/updatePublicPrivateOperator").blockingHandler(auth.handle((ctx) -> {
```
aulme, line 99: URLs should not use capitals - different web servers handle casing differently so it tends to create confusion. The convention tends to be to use dashes instead.

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
sunnywu: Changed as suggested

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
sunnywu: it was my lazy copy&paste off old code - i think once upon a time there was just ClientKey hence it was 'c' and at some point OperatorKey was created afterwards...

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
sunnywu: noted have removed this and just re-use the good old /update end point altogether.

PR #23, opened 2023-01-03
UID2-602 Add roles APIs to operator key

jingyi-gao-ttd:


pom.xml
```diff
@@ -16,7 +16,7 @@
         <!-- check micrometer.version vertx-micrometer-metrics consumes before bumping up -->
         <micrometer.version>1.1.0</micrometer.version>
         <junit-jupiter.version>5.7.0</junit-jupiter.version>
-        <uid2-shared.version>1.2.1</uid2-shared.version>
+        <uid2-shared.version>1.11.0</uid2-shared.version>
```
gmsdelmundo-ttd, line 19: Don't forget to update this version once you build out the new uid2-shared release

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
```diff
@@ -144,22 +169,48 @@ private void handleOperatorAdd(RoutingContext rc) {
...
+            final String protocol = rc.queryParams().contains("protocol")
+                    ? RequestUtil.validateOperatorProtocol(rc.queryParam("protocol").get(0))
+                    : null;
             if (protocol == null) {
                 ResponseUtil.error(rc, 400, "no protocol specified");
                 return;
             }
 
-            List<OperatorKey> operators = this.operatorKeyProvider.getAll()
+            Set<Role> roles = RequestUtil.getRoles(rc.queryParam("roles").get(0));
```
gmsdelmundo-ttd, line 194: `rc.queryParam("roles").get(0)` may throw an NPE especially if you want to support backwards compatibility.

It would be better to do `rc.queryParam("roles") == null ? Set.of() : rc.queryParam("roles").get(0)`

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
```diff
@@ -298,4 +386,41 @@ private void handleOperatorRekey(RoutingContext rc) {
...
+            Optional<OperatorKey> existingOperator = this.operatorKeyProvider.getAll()
+                    .stream().filter(o -> o.getName().equals(name))
+                    .findFirst();
+            if (!existingOperator.isPresent()) {
+                ResponseUtil.error(rc, 404, "operator key not found");
+                return;
+            }
+
+            Set<Role> roles = RequestUtil.getRoles(rc.queryParam("roles").get(0));
+            if (roles == null) {
```
gmsdelmundo-ttd, line 405: Same comment here about NPE on roles

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
gmsdelmundo-ttd: Also, you need to validate whether all roles are not null

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
gmsdelmundo-ttd: Same comment here about null role validation

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
```diff
@@ -377,4 +386,41 @@ private void handleOperatorRekey(RoutingContext rc) {
...
+                ResponseUtil.error(rc, 400, "incorrect or none roles specified");
```
gmsdelmundo-ttd, line 406: Error message can be "No roles specified"

src/test/java/com/uid2/admin/vertx/OperatorKeyServiceTest.java
```diff
@@ -32,18 +37,29 @@ private void checkOperatorKeyResponse(OperatorKey[] expectedOperators, Object[]
             assertEquals(expectedOperator.getProtocol(), actualOperator.getString("protocol"));
             assertEquals(expectedOperator.isDisabled(), actualOperator.getBoolean("disabled"));
             assertEquals(expectedOperator.getSiteId(), actualOperator.getInteger("site_id"));
+
+            List<Role> actualRoles = actualOperator.getJsonArray("roles").stream()
+                    .map(r -> Role.valueOf((String) r))
+                    .collect(Collectors.toList());
+            assertEquals(expectedOperator.getRoles().size(), actualRoles.size());
+            for (Role role : expectedOperator.getRoles()) {
+                assertTrue(actualRoles.contains(role));
```
gmsdelmundo-ttd, line 46: You can simplify this comparison: `assertTrue(actualRoles.containsAll(expectedOperator.getRoles()))`

src/test/java/com/uid2/admin/vertx/OperatorKeyServiceTest.java
```diff
@@ -62,14 +78,21 @@ void operatorAdd(Vertx vertx, VertxTestContext testContext) {
     @Test
     void operatorAddUnknownSiteId(Vertx vertx, VertxTestContext testContext) {
         fakeAuth(Role.OPERATOR_MANAGER);
-        post(vertx, "api/operator/add?name=test_client&protocol=trusted&site_id=5", "", expectHttpError(testContext, 400));
+        post(vertx, "api/operator/add?name=test_client&protocol=trusted&site_id=5&roles=optout", "", expectHttpError(testContext, 400));
+    }
+
+    @Test
+    void operatorKeyAddInvalidRole(Vertx vertx, VertxTestContext testContext) {
```
gmsdelmundo-ttd, line 85: After adding the null role validation, add another test case that includes one valid role and one null role

src/test/java/com/uid2/admin/vertx/OperatorKeyServiceTest.java
```diff
@@ -32,18 +37,29 @@ private void checkOperatorKeyResponse(OperatorKey[] expectedOperators, Object[]
...
     @Test
     void operatorAdd(Vertx vertx, VertxTestContext testContext) {
         fakeAuth(Role.OPERATOR_MANAGER);
+        Set<Role> roles = new HashSet<>();
+        roles.add(Role.OPTOUT);
+        roles.add(Role.OPERATOR);
         setSites(new Site(5, "test_site", true));
         OperatorKey[] expectedOperators = {
-                new OperatorKey("", "test_operator", "test_operator", "trusted", 0, false, 5)
+                new OperatorKey("", "test_operator", "test_operator", "trusted", 0, false, 5).withRoles(roles)
```
gmsdelmundo-ttd, line 59: You could use your new OperatorKey constructor that has `roles` as an input

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
```diff
@@ -184,6 +187,11 @@ private void handleOperatorAdd(RoutingContext rc) {
                 return;
             }
 
+            Set<Role> roles = RequestUtil.getRoles(rc.queryParam("roles").get(0)) == null ? Set.of() : RequestUtil.getRoles(rc.queryParam("roles").get(0));
+            if (roles.stream().allMatch(Objects::isNull)) {
+                ResponseUtil.error(rc, 400, "No roles specified");
```
sunnywu, line 192: so when a user wants to add a new Operator, the OPERATOR role must be explicitly specified in the textbox or otherwise it will hit this error here? Is that what we want to do (seems a bit cumbersome)? 

If we keep it this way, the error msg should say something like "No roles specified - at least OPERATOR role must be specified here" to clarify what the users need to input

jingyi-gao-ttd: hmmm, i guess with add new operator we should be able to leave the role to be empty, I will remove the check here

src/test/java/com/uid2/admin/vertx/OperatorKeyServiceTest.java
```diff
@@ -62,14 +76,47 @@ void operatorAdd(Vertx vertx, VertxTestContext testContext) {
...
+        fakeAuth(Role.OPERATOR_MANAGER);
+        post(vertx, "api/operator/add?name=test_client&protocol=trusted&site_id=5&roles=nonexist", "", expectHttpError(testContext, 400));
     }
 
     @Test
     void operatorUpdate(Vertx vertx, VertxTestContext testContext) {
         fakeAuth(Role.ADMINISTRATOR);
         setSites(new Site(5, "test_site", true));
-        setOperatorKeys(new OperatorKey("", "test_operator", "test_operator", "trusted", 0, false, null));
+        setOperatorKeys(new OperatorKey("", "test_operator", "test_operator", "trusted", 0, false));
```
sunnywu, line 118: you probably shouldn't use OperatorKey constructors without Role param any more - those are for backward compatibility only and going forward OperatorKey must have role

sunnywu:
Can't quite add the comment inside the unchanged method -  should you allow modifying roles in com.uid2.admin.vertx.service.OperatorKeyService#handleOperatorUpdate as well?

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
```diff
@@ -184,6 +187,13 @@ private void handleOperatorAdd(RoutingContext rc) {
                 return;
             }
 
+            Set<Role> roles = RequestUtil.getRoles(rc.queryParam("roles").get(0)) == null ? new HashSet<>() : RequestUtil.getRoles(rc.queryParam("roles").get(0));
```
sunnywu, line 190: actually when i was testing my changes, i noticed that "rc.queryParam("roles").get(0)" will throw exception - what you want to check is actually "rc.queryParam("roles").isEmpty()" as queryParams return a list. If returned list is empty, that param doesn't exist so should create a new HashSet<>() instead. might want to fix this in other places within this file

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
```diff
@@ -184,6 +187,18 @@ private void handleOperatorAdd(RoutingContext rc) {
                 return;
             }
 
+            Set<Role> roles = (!rc.queryParams().contains("roles")
+                    || RequestUtil.getRoles(rc.queryParam("roles").get(0)) == null)
+                        ? new HashSet<>()
+                        : RequestUtil.getRoles(rc.queryParam("roles").get(0));
+
+            if (rc.queryParams().contains("roles")
+                    && rc.queryParam("roles").get(0) != ""
```
gmsdelmundo-ttd, line 196: You can use `!rc.queryParam("roles").get(0).trim().isEmpty()`

`.isBlank()` works as well, but it's a Java 11 feature

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
```diff
@@ -184,6 +187,18 @@ private void handleOperatorAdd(RoutingContext rc) {
...
+                    && roles.stream().allMatch(Objects::isNull)) {
```
gmsdelmundo-ttd, line 197: This one should be `anyMatch()` instead of `allMatch()`

jingyi-gao-ttd: I am thinking if user is given multiple roles and there is one invalid role, should we throw error or ignore it?

src/test/java/com/uid2/admin/vertx/OperatorKeyServiceTest.java
```diff
@@ -93,6 +167,55 @@ void operatorUpdate(Vertx vertx, VertxTestContext testContext) {
...
+            } catch (Exception ex) {
+                fail(ex);
+            }
+
+            testContext.completeNow();
+        });
+    }
+
+    @Test
+    void operatorKeySetInvalidRole(Vertx vertx, VertxTestContext testContext) {
```
gmsdelmundo-ttd, line 202: Could you add one more test case that has two roles? One role is valid, another role is null/empty

gmsdelmundo-ttd (approved):

PR #28, opened 2023-01-11
Created job dispatcher for syncing site-specific data

gmsdelmundo-ttd:


src/main/java/com/uid2/admin/job/JobDispatcher.java
```diff
@@ -0,0 +1,118 @@
...
+            currentJob = JOB_QUEUE.poll();
+            assert currentJob != null;
+            currentJobId = currentJob.getId();
+            LOGGER.info("Executing job: {} ({} jobs remaining in queue)", currentJobId, JOB_QUEUE.size());
+        }
+
+        JOB_EXECUTOR.execute(() -> {
+            for (int retryCount = 1; retryCount <= maxRetries; retryCount++) {
+                try {
+                    currentJob.execute();
```
sunnywu, line 99: should we log when a job is finished successfully?

src/test/java/com/uid2/admin/job/JobDispatcherTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Test
+    public void testJobExecutionWithOneJob() throws Exception {
+        jobDispatcher.enqueue(new TestJob());
+        jobDispatcher.start(10, 3);
+        Thread.sleep(20);
+
+        assertEquals(1, executionCount.get());
+    }
+
```
sunnywu, line 87: could you add a test that test adding multiple different unique TestJobs and still asserts:
```
assertEquals(X, executionCount.get());
assertTrue(jobDispatcher.getJobQueueInfo().isEmpty());
```
?

src/main/java/com/uid2/admin/job/model/JobInfo.java
```diff
@@ -0,0 +1,29 @@
...
+    private final boolean executing;
+    private final Instant executionTime;
+
+    public JobInfo(String id, boolean executing) {
+        this.id = id;
+        this.executing = executing;
+        this.executionTime = Instant.now();
+    }
+
+    public String getId() {
```
sunnywu, line 17: off topic: should the job id of our use case be something like "Timestamp-SiteId(-DataType-GlobalMetadataVersionGeneratedFrom)"?

src/test/java/com/uid2/admin/job/JobDispatcherTest.java
```diff
@@ -0,0 +1,109 @@
...
+    @Test
+    public void testJobExecutionWithDuplicateJobs() throws Exception {
+        jobDispatcher.enqueue(new TestJob());
+        jobDispatcher.enqueue(new TestJob());
+        jobDispatcher.start(10, 3);
+        Thread.sleep(20);
```
sunnywu, line 93: in my historical experience, the CI build box performance varies so at times it could take longer than expected to trigger "alarm" and it's possible it takes longer than 20ms to run the job. So tend to not use a real alarm when writing unit test. 

given how simple this test is (and CI load of our UID2 project) maybe it won't fail due to this reason

src/main/java/com/uid2/admin/job/model/JobInfo.java
gmsdelmundo-ttd: I think this can be implemented as an abstract class that implements the interface for sync jobs

i.e. `public abstract class SyncJob implements Job` and the job ID you mentioned can be part of the constructor?

sunnywu: yeh i think we should identify the job easily by forming the job id with info above cc @aulme

sunnywu (approved):

PR #33, opened 2023-01-17
Use `Clock` class instead of static `Instant.now()`

cYKatherine:


src/test/java/com/uid2/admin/vertx/EncryptionKeyServiceTest.java
```diff
@@ -23,17 +25,24 @@ public class EncryptionKeyServiceTest extends ServiceTestBase {
...
     private static final int SITE_KEY_ACTIVATES_IN_SECONDS = 36000;
     private static final int SITE_KEY_EXPIRES_AFTER_SECONDS = 72000;
-
+    private Clock clock = mock(Clock.class);
     private EncryptionKeyService keyService = null;
 
+    @BeforeEach
+    void setUp() {
+        // seconds since epoch, DO NOT set this to 1 as it will make the time before 1970-00-00 as it will screw up the test...
+        Long now = 100000L;
```
sunnywu, line 34: should we set a more realistic time here?

cYKatherine: Set it to be one year 👍

cYKatherine: Actually, just change to `key creation time + 10 days` after the discussion with @aulme . Due to my change, if we use a fix timestamp for current time, it will make key creation time go backwards and I would like to avoid that.

thomasm-ttd (approved):

PR #35, opened 2023-01-18
UID2-599 Filter out keys duration rotation

cYKatherine:


src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -68,8 +74,11 @@ public EncryptionKeyService(JsonObject config,
 
         masterKeyActivatesIn = Duration.ofSeconds(config.getInteger(MASTER_KEY_ACTIVATES_IN_SECONDS));
         masterKeyExpiresAfter = Duration.ofSeconds(config.getInteger(MASTER_KEY_EXPIRES_AFTER_SECONDS));
+        masterKeyRotationCutoffTime = Duration.ofDays(config.getInteger(MASTER_KEY_ROTATION_CUT_OFF_DAYS));
```
thomasm-ttd, line 77: What happens if these are not specified in config? Does it return 0 or throw an exception? If it does return 0, what will the code do?

src/test/java/com/uid2/admin/vertx/EncryptionKeyServiceTest.java
```diff
@@ -172,9 +260,9 @@ void rotateMasterKeyNewEnough(Vertx vertx, VertxTestContext testContext) throws
 
         final EncryptionKey[] keys = {
                 new EncryptionKey(11, null, Instant.ofEpochMilli(KEY_CREATE_TIME_IN_MILLI), Instant.ofEpochMilli(KEY_ACTIVATE_TIME_IN_MILLI), Instant.ofEpochMilli(KEY_EXPIRE_TIME_IN_MILLI), -1),
-                new EncryptionKey(12, null, Instant.ofEpochMilli(KEY_CREATE_TIME_IN_MILLI+1), clock.now().plusSeconds(MASTER_KEY_ACTIVATES_IN_SECONDS + 1000), Instant.MAX, -1),
+                new EncryptionKey(12, null, Instant.ofEpochMilli(KEY_CREATE_TIME_IN_MILLI+1), clock.now().plusSeconds(MASTER_KEY_ACTIVATES_IN_SECONDS + TEN_DAYS_IN_SECONDS * 10), clock.now().plusSeconds(MASTER_KEY_EXPIRES_AFTER_SECONDS + TEN_DAYS_IN_SECONDS * 10), -1),
```
sunnywu, line 263: TEN_DAYS_IN_SECONDS is already defined as 864000L (86400 a day * 10 days), why does it * 10 again?

sunnywu: or if you want 100 days might be better to define A_HUNDRED_DAYS_IN_SECONDS for readability?

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
sunnywu: to be safe better set a huge default value in code if it can't be read from config

src/test/java/com/uid2/admin/vertx/EncryptionKeyServiceTest.java
cYKatherine: You are right! I will make that change.

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -137,7 +146,8 @@ private EncryptionKey addSiteKey(int siteId, Duration activatesIn) throws Except
         // force refresh manually
         this.keyProvider.loadContent();
 
-        return addSiteKeys(Arrays.asList(siteId), activatesIn, siteKeyExpiresAfter).get(0);
+        // Set cutoffTime to be 0 here as we don't want to filter out the keys that have passed the cutoff time
+        return addSiteKeys(Arrays.asList(siteId), activatesIn, siteKeyExpiresAfter, Duration.ofDays(0)).get(0);
```
sunnywu, line 150: maybe my understanding is wrong - i take it you mean you wanna KEEP ALL the keys regardless the cutoff time config or is this wrong? if you set the cutoff time to 0, as soon as the key expires it will be filtered out in the code below right?

```
 final List<EncryptionKey> keys = this.keyProvider.getSnapshot().getActiveKeySet().stream()
                .sorted(Comparator.comparingInt(EncryptionKey::getId))
                .filter(k -> now.compareTo(k.getExpires().plus(cutoffTime.toDays(), ChronoUnit.DAYS)) < 0)
                .collect(Collectors.toList());
```

cYKatherine: You are absolutely right... I will make the change. 

I'm surprised we don't have tests to cover this 🤔

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
sunnywu: you can use getInteger(String key, Integer def) method to set a default value if config didn't specify a value

sunnywu:
so in general if we wanna disable the filtering, do we just set the cutoff time config to maximum (9999999999)? it might be good idea to create a config/feature switch to indicate if we wanna enable/disable this feature (in case it goes wrong)

would be good to have a rollback plan in case of things go wrong - e.g. worst case scenario is to revert to a backup file for example. (just having an idea what needs to be done is good enough, not need to write 'a proper plan')

cYKatherine:
> so in general if we wanna disable the filtering, do we just set the cutoff time config to maximum (9999999999)? it might be good idea to create a config/feature switch to indicate if we wanna enable/disable this feature (in case it goes wrong)
> 
> would be good to have a rollback plan in case of things go wrong - e.g. worst case scenario is to revert to a backup file for example. (just having an idea what needs to be done is good enough, not need to write 'a proper plan')

cYKatherine:
> so in general if we wanna disable the filtering, do we just set the cutoff time config to maximum (9999999999)? it might be good idea to create a config/feature switch to indicate if we wanna enable/disable this feature (in case it goes wrong)
> 
> would be good to have a rollback plan in case of things go wrong - e.g. worst case scenario is to revert to a backup file for example. (just having an idea what needs to be done is good enough, not need to write 'a proper plan')

Added a flag in config file and set it default to be false :)

src/test/java/com/uid2/admin/vertx/EncryptionKeyServiceTest.java
```diff
@@ -154,11 +188,131 @@ void rotateMasterKey(Vertx vertx, VertxTestContext testContext) throws Exception
...
+            } catch (Exception ex) {
+                fail(ex);
+            }
+            testContext.completeNow();
+        });
+    }
+
+    @Test
+    void filterOutRefreshKeysOverCutoffTimeWithFlagOn(Vertx vertx, VertxTestContext testContext) throws Exception {
+        this.config.put("filter_key_over_cut_off_days", true);
```
cYKatherine, line 265: This part is nasty I know as it repeated 3 times, I should've put it in a function :( 

But we will be removing this flag later on if everything is okay. So I suggest to keep it this way and remove it later on.

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -316,22 +332,24 @@ private RotationResult rotateKeys(Predicate<Integer> siteSelector, Duration minA
...
     }
 
-    private List<EncryptionKey> addSiteKeys(Iterable<Integer> siteIds, Duration activatesIn, Duration expiresAfter)
+    private List<EncryptionKey> addSiteKeys(Iterable<Integer> siteIds, Duration activatesIn, Duration expiresAfter, Duration cutoffTime)
             throws Exception {
+        final Instant now = clock.now();
+
         final List<EncryptionKey> keys = this.keyProvider.getSnapshot().getActiveKeySet().stream()
                 .sorted(Comparator.comparingInt(EncryptionKey::getId))
+                .filter(filterKeyOverCutOffTime ? k -> now.compareTo(k.getExpires().plus(cutoffTime.toDays(), ChronoUnit.DAYS)) < 0 : k -> true)
```
cYKatherine, line 346: This is where the feature flag/switch comes in. We will only filter keys out when we set `filterKeyOverCutOffTime` to be true.

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -137,7 +149,8 @@ private EncryptionKey addSiteKey(int siteId, Duration activatesIn) throws Except
         // force refresh manually
         this.keyProvider.loadContent();
 
-        return addSiteKeys(Arrays.asList(siteId), activatesIn, siteKeyExpiresAfter).get(0);
+        // Set cutoffTime to be maximum here as we don't want to filter out the keys that have passed the cutoff time
+        return addSiteKeys(Arrays.asList(siteId), activatesIn, siteKeyExpiresAfter, Duration.ofNanos(Instant.MAX.getNano())).get(0);
```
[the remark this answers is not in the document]
sunnywu: this is fine but strange getMillie() overflows...

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -157,11 +170,14 @@ private void handleKeyList(RoutingContext rc) {
 
     private void handleRotateMasterKey(RoutingContext rc) {
         try {
-            final RotationResult result = rotateKeys(rc, masterKeyActivatesIn, masterKeyExpiresAfter,
-                    s -> s == Const.Data.MasterKeySiteId || s == Const.Data.RefreshKeySiteId);
+            final RotationResult masterKeyResult = rotateKeys(rc, masterKeyActivatesIn, masterKeyExpiresAfter, masterKeyRotationCutoffTime,
+                s -> s == Const.Data.MasterKeySiteId);
+            final RotationResult refreshKeyResult = rotateKeys(rc, masterKeyActivatesIn, masterKeyExpiresAfter, refreshKeyRotationCutOffTime,
+                    s -> s == Const.Data.RefreshKeySiteId);
```
sunnywu, line 176: i'm not too familiar with the rotate codes but why do we need to do this second rotateKeys call ofr RefreshSiteId now? or why we didn't need to do this before?

sunnywu:
do you wanna add your new config variables with default values to the uid2-admin\conf\local-config.json for testing?

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
aulme: Good point - are some keys that never got rotated before starting to get rotated? Didn't think there's such component to this story.

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
aulme: I think it would be better to replace the lambda with a method reference so we can have a descriptive name here. It takes quite a bit of brainwork to figure out what it actually does.
Probably also better to just have the function return true in a guard clause than have a ternary returning an always true lambda.

src/test/java/com/uid2/admin/vertx/EncryptionKeyServiceTest.java
aulme: Fine for code we know we will remove.

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
aulme: To think of it, probably worth extracting the entire filtering functionality into a separate collaborator class/method. This file is already having way too much in it, it wouldn't make sense for your PR to fix all of that of course, but perhaps we can stop adding to the problem. Then we could also add unit tests rather than rely solely on rather lengthy and slow integration tests for everything. Happy to explain what I mean if you need help.

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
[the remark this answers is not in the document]
cYKatherine: There is no `Instant.toYear` or `Instant.toDate` method, converting 1k years to seconds might not be ideal :(

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
cYKatherine: The original function does `rotateKeys(rc, masterKeyActivatesIn, masterKeyExpiresAfter, s -> s == Const.Data.MasterKeySiteId || s -> s == Const.Data.RefreshKeySiteId);` which includes both master key and refresh key. 

The reason why I'm doing it twice is because we need to pass in both `masterKeyRotationCutoffTime` and `refreshKeyRotationCutoffTime` to this function and they have different values.

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -157,11 +170,14 @@ private void handleKeyList(RoutingContext rc) {
 
     private void handleRotateMasterKey(RoutingContext rc) {
```
thomasm-ttd, line 175: This will now return the keys in a different order - it will return all the MasterKeySite first, and then all the RefreshKeySite second (in the same json file). Does this matter? I don't know who calls it, so not sure

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
sunnywu: ah i see!

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
sunnywu: i dont think it matters as it's just a unsorted list in json. @aulme any view?

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
sunnywu, line 180: so each rotateKeys call will indirectly call s3 upload once so essentially if you call rotateKeys twice you are doing 2 updates/uploads for encryption key files. it doesn't really matter as it's called within a single writeLock so there won't be inconsistency issue i think (e.g. master is updated but not refresh keys yet and an operator is trying to download them all). but it's an inefficiency. if you are in a hurry i suggest just create a jira ticket for future improvement.

or if you wanna fix it maybe you can create a Predicate CutoffTimeSelector (similar to the Predicate<Integer> siteSelector) that will take in siteId and return cutoff time accordingly.

scong-ttd: there will be a race condition over max_id (metadata in memory is updated by downloading the new metadata file I suppose, so continuous writing without fetching will create keys with same id and conflict)

sunnywu: best back to having 1 rotateKeys call then

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -358,4 +371,23 @@ private JsonObject toJson(EncryptionKey key) {
...
+        int siteId = 8;
+        Duration cutoffTime;
+        switch (siteId) {
+            case -1:  cutoffTime = masterKeyRotationCutoffTime;
+                break;
+            case -2:  cutoffTime = refreshKeyRotationCutOffTime;
+                break;
+            default: cutoffTime = siteKeyRotationCutOffTime;
+                break;
+        }
```
sunnywu, line 385: can you refactor to reduce the (extra but minor) processing such as:

```
if (!(filterKeyOverCutOffTime && duringRotation)) {
    return true;
}

//rest of switch and compareTo statement
....
```

cYKatherine: done

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -358,4 +371,24 @@ private JsonObject toJson(EncryptionKey key) {
...
-}
+
+    private boolean isWithinCutOffTime(EncryptionKey key, Instant now, boolean duringRotation) {
+        if (!(filterKeyOverCutOffTime && duringRotation)) {
+            return true;
+        }
+        int siteId = key.getSiteId();
+        Duration cutoffTime;
+        switch (siteId) {
+            case -1:
```
sunnywu, line 382: sorry forgot to mention should use Const.Data.MasterKeySiteId /Const.Data.RefreshKeySiteId here instead of hardcoded numbers

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
aulme: I guess it's fine to not preserve order if we don't care about (which I think we don't). Lists in JSON are ordered however.

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -358,4 +371,23 @@ private JsonObject toJson(EncryptionKey key) {
         jo.put("expires", key.getExpires().toEpochMilli());
         return jo;
     }
-}
+
+    private boolean isWithinCutOffTime(EncryptionKey key, Instant now, boolean duringRotation) {
+        int siteId = 8;
```
thomasm-ttd: Why is the siteId fixed? What site is siteId 8? Is it a special site?

conf/local-config.json
```diff
@@ -20,9 +20,13 @@
   "admin_key_prefix": "ADINT",
   "master_key_activates_in_seconds": 30,
   "master_key_expires_after_seconds": 60000,
+  "master_key_rotation_cut_off_days": 37,
```
scong-ttd, line 23: Why 37?

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -157,11 +169,11 @@ private void handleKeyList(RoutingContext rc) {
 
     private void handleRotateMasterKey(RoutingContext rc) {
         try {
-            final RotationResult result = rotateKeys(rc, masterKeyActivatesIn, masterKeyExpiresAfter,
-                    s -> s == Const.Data.MasterKeySiteId || s == Const.Data.RefreshKeySiteId);
+            final RotationResult masterKeyResult = rotateKeys(rc, masterKeyActivatesIn, masterKeyExpiresAfter,
+                s -> s == Const.Data.MasterKeySiteId || s == Const.Data.RefreshKeySiteId);
```
scong-ttd, line 173: then master key and refresh key will have the same expiry?

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -316,22 +328,24 @@ private RotationResult rotateKeys(Predicate<Integer> siteSelector, Duration minA
...
         }
 
-        result.rotatedKeys = addSiteKeys(siteIds, activatesIn, expiresAfter);
+        result.rotatedKeys = addSiteKeys(siteIds, activatesIn, expiresAfter, true);
 
         return result;
     }
 
-    private List<EncryptionKey> addSiteKeys(Iterable<Integer> siteIds, Duration activatesIn, Duration expiresAfter)
+    private List<EncryptionKey> addSiteKeys(Iterable<Integer> siteIds, Duration activatesIn, Duration expiresAfter, boolean duringRotation)
```
aulme, line 336: NP it's a convention to name boolean vars with is/are e.g. isDuringRotation so it's obvious which way the boolean goes.

conf/local-config.json
cYKatherine: https://atlassian.thetradedesk.com/confluence/display/EN/UID2-599+Delete+keys+that+has+expired+for+more+than+30+days+during+key+rotation

Here is the detailed explanation! 

Short version is:

- refresh key cut off = expiry timestamp + refresh token lifetime (30d) + troubleshooting buffer (7d)
- master key cut off = expiry timestamp + max advertising token lifetime (30d) + troubleshooting buffer (7d)
- site key cut off = expiry timestamp + max advertising token lifetime (30d) + troubleshooting buffer (7d)

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
cYKatherine: Yes, that's the original logic and idk if we plan to change that.

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
cYKatherine: Changed!

aulme (approved):

sunnywu (approved):

conf/local-config.json
scong-ttd: Uh. The page has some mistakes (like refresh_key lifetime is definitely not 1h) but your analysis remains correct. So 30d is the estimated maximum lifetime given by UID Sharing task force. Thanks!

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
scong-ttd: Just checking if it's intentional. Good.

scong-ttd (approved):

PR #39, opened 2023-01-20
UID2-650 Added APIs to update metadata version in the metadata files without changing the underlying data files

aulme:


sunnywu:
Testing locally on Localstack s3 metadata files and can confirm the metadata files of client keys/encryption keys and encryption key acls are modified and only the version numbers have been updated and nothing else. 

also created a admin UI Overwrite metadata button to trigger this in 1 go.

PR #41, opened 2023-01-23
Add API endpoint for adding site key

mcollins-ttd:


src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -139,6 +146,43 @@ private void handleRotateMasterKey(RoutingContext rc) {
         }
     }
 
+    private void handleAddSiteKey(RoutingContext rc) {
+        final Optional<Integer> siteIdOpt = RequestUtil.getSiteId(rc, "site_id");
+        if (!siteIdOpt.isPresent()) {
+            return;
```
atarassov-ttd, line 152: Why is site id optional and what happens if you don't supply it?

webroot/adm/encryption-key.html
```diff
@@ -55,6 +56,13 @@ <h3>Output</h3>
             doApiCall('POST', url, '#standardOutput', '#errorOutput');
         });
 
+        $('#doAddSiteKey').on('click', function () {
+            var siteId = encodeURIComponent($('#siteId').val());
+            var url = '/api/key/add?&site_id=' + siteId;
```
atarassov-ttd, line 61: nit: `?site_id`

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
mcollins-ttd: If there is no site ID supplied - or it can't be parsed as an int - `RequestUtils.getSiteId` will fail the request with a 400.

It might be good to rename `getSiteId` so that it indicates that it will fail the request.

webroot/adm/encryption-key.html
mcollins-ttd: Fixed, thanks!

atarassov-ttd (approved):

PR #49, opened 2023-01-31
Added execution duration metric and stats API for job dispatcher

gmsdelmundo-ttd:
Testing:

- Added unit tests
- While a job was executing, checked http://localhost:9089/metrics and saw the job execution duration in the output

### DIFF
--- a/src/main/java/com/uid2/admin/Main.java
+++ b/src/main/java/com/uid2/admin/Main.java
@@ -135,7 +135,7 @@ public class Main {
             WriteLock writeLock = new WriteLock();
             IKeyGenerator keyGenerator = new SecureKeyGenerator();
             ISaltRotation saltRotation = new SaltRotation(config, keyGenerator);
-            JobDispatcher jobDispatcher = new JobDispatcher("dispatcher", 1000 * 60, 3, clock);
+            JobDispatcher jobDispatcher = new JobDispatcher("job-dispatcher", 1000 * 60, 3, clock);
 
             final EncryptionKeyService encryptionKeyService = new EncryptionKeyService(
                     config, auth, writeLock, encryptionKeyStoreWriter, keyProvider, keyGenerator, clock);
--- a/src/main/java/com/uid2/admin/Main.java
+++ b/src/main/java/com/uid2/admin/Main.java
@@ -6,6 +6,7 @@ import com.uid2.admin.auth.GithubAuthHandlerFactory;
 import com.uid2.admin.auth.IAuthHandlerFactory;
 import com.uid2.admin.job.JobDispatcher;
 import com.uid2.admin.job.jobsync.PrivateSiteDataSyncJob;
+import com.uid2.admin.job.model.Job;
 import com.uid2.admin.model.Site;
 import com.uid2.admin.secret.IKeyGenerator;
 import com.uid2.admin.secret.ISaltRotation;
@@ -134,6 +135,7 @@ public class Main {
             WriteLock writeLock = new WriteLock();
             IKeyGenerator keyGenerator = new SecureKeyGenerator();
             ISaltRotation saltRotation = new SaltRotation(config, keyGenerator);
+            JobDispatcher jobDispatcher = new JobDispatcher("dispatcher", 1000 * 60, 3, clock);
 
             final EncryptionKeyService encryptionKeyService = new EncryptionKeyService(
                     config, auth, writeLock, encryptionKeyStoreWriter, keyProvider, keyGenerator, clock);
@@ -148,7 +150,8 @@ public class Main {
                     new SaltService(auth, writeLock, saltStoreWriter, saltProvider, saltRotation),
                     new SiteService(auth, writeLock, siteStoreWriter, siteProvider, clientKeyProvider),
                     new PartnerConfigService(auth, writeLock, partnerStoreWriter, partnerConfigProvider),
-                    new PrivateSiteDataRefreshService(auth, writeLock, config),
+                    new PrivateSiteDataRefreshService(auth, jobDispatcher, writeLock, config),
+                    new JobDispatcherService(auth, jobDispatcher)
             };
 
             AdminVerticle adminVerticle = new AdminVerticle(config, authHandlerFactory, auth, adminUserProvider, services);
@@ -161,10 +164,9 @@ public class Main {
 
             //UID2-575 setup a job dispatcher that will write private site data periodically if there is any changes
             //check job for every minute
-            JobDispatcher.getInstance().start(1000 * 60, 3);
             PrivateSiteDataSyncJob job = new PrivateSiteDataSyncJob(config, writeLock);
-            JobDispatcher.getInstance().enqueue(job);
-            JobDispatcher.getInstance().executeNextJob(3);
+            jobDispatcher.enqueue(job);
+            jobDispatcher.executeNextJob();
         } catch (Exception e) {
             LOGGER.fatal("failed to initialize core verticle", e);
             System.exit(-1);
--- a/src/main/java/com/uid2/admin/job/JobDispatcher.java
+++ b/src/main/java/com/uid2/admin/job/JobDispatcher.java
@@ -77,7 +77,7 @@ public class JobDispatcher {
                     && JOB_QUEUE.stream().noneMatch(queuedJob -> queuedJob.getId().equals(id))) {
                 LOGGER.info("Queueing new job: {}", id);
                 JOB_QUEUE.add(job);
-                job.setEnqueueTime(clock.now());
+                job.setAddedToQueueAt(clock.now());
             } else {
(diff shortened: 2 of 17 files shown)


src/main/java/com/uid2/admin/Main.java
```diff
@@ -161,10 +164,9 @@ public void run() {
 
             //UID2-575 setup a job dispatcher that will write private site data periodically if there is any changes
             //check job for every minute
-            JobDispatcher.getInstance().start(1000 * 60, 3);
             PrivateSiteDataSyncJob job = new PrivateSiteDataSyncJob(config, writeLock);
-            JobDispatcher.getInstance().enqueue(job);
-            JobDispatcher.getInstance().executeNextJob(3);
+            jobDispatcher.enqueue(job);
+            jobDispatcher.executeNextJob();
```
aulme, line 169: Why is this changed? Not sure why it did 3 before either though.

src/main/java/com/uid2/admin/job/JobDispatcher.java
```diff
@@ -2,39 +2,56 @@
...
+            int maxRetries,
+            Clock clock) {
+        this.id = id;
+        this.intervalMs = intervalMs;
+        this.maxRetries = maxRetries;
+        this.clock = clock;
+
+        Gauge.builder("uid2.job_dispatcher.execution_duration_ms", this::getExecutionTime)
+                .tag("job_dispatcher", id)
+                .description("gauge for " + id + " execution time")
```
aulme, line 45: Don't need to include "gauge for", just like variable names don't need the word "variable" in them :)

src/main/java/com/uid2/admin/job/JobDispatcher.java
```diff
@@ -141,9 +134,47 @@ private void run(int maxRetries) {
...
+    public boolean isExecutingJob() {
+        return currentJob != null;
+    }
+
+    public boolean isStarted() {
+        return started;
+    }
+
+    private long getExecutionTime() {
+        synchronized (jobLock) {
```
aulme, line 176: I suggest wee don't use lock here - prom will call this very frequently and will keep the lock busy.

src/main/java/com/uid2/admin/job/JobDispatcher.java
```diff
@@ -141,9 +134,47 @@ private void run(int maxRetries) {
...
+
+        synchronized (jobLock) {
+            if (isExecutingJob()) {
+                jobInfos.add(new JobInfo(currentJob, true));
+            }
+            jobInfos.addAll(jobQueue.stream()
+                    .map(job -> new JobInfo(job, false))
+                    .collect(Collectors.toList()));
+        }
+        return jobInfos;
```
aulme, line 154: We need tests for this

src/main/java/com/uid2/admin/job/JobDispatcher.java
```diff
@@ -141,9 +134,47 @@ private void run(int maxRetries) {
...
+        return currentJob != null;
+    }
+
+    public boolean isStarted() {
+        return started;
+    }
+
+    private long getExecutionTime() {
+        synchronized (jobLock) {
+            return currentJob == null ? 0 : ChronoUnit.MILLIS.between(currentJob.getStartedExecutingAt(), Instant.now());
```
aulme, line 177: We need tests for this

src/main/java/com/uid2/admin/Main.java
gmsdelmundo-ttd: changing it from singleton to non-singleton

3 was there because you had to pass in `maxRetries`

src/main/java/com/uid2/admin/job/JobDispatcher.java
gmsdelmundo-ttd: I think all the other descriptions have the same format so I'm just following those descriptions

src/main/java/com/uid2/admin/job/JobDispatcher.java
gmsdelmundo-ttd: Sure will remove it

src/main/java/com/uid2/admin/job/JobDispatcher.java
gmsdelmundo-ttd: Actually will update this a bit, will add a relevant test case after

src/main/java/com/uid2/admin/job/JobDispatcher.java
gmsdelmundo-ttd: already have tests for this

src/main/java/com/uid2/admin/Main.java
aulme: ah right

aulme (approved):

src/main/java/com/uid2/admin/job/JobDispatcher.java
```diff
@@ -2,39 +2,55 @@
...
+            Clock clock) {
+        this.id = id;
+        this.intervalMs = intervalMs;
+        this.maxRetries = maxRetries;
+        this.clock = clock;
+
+        Gauge.builder("uid2.job_dispatcher.execution_duration_ms", this::getExecutionDuration)
+                .tag("job_dispatcher", id)
+                .description("gauge for " + id + " execution time")
+                .register(Metrics.globalRegistry);
```
sunnywu, line 45: so prometheus will poll for this execution_duration_ms value periodically via the Gauge mechanism? would alert be based on such gauge? (e.g. if prometheus always poll at the time when currnetJob is null then this Gauge wouldn't be helpful? ) 

or should we always store the mnost recent job's executionDuration in JobDispatcher and getExecutionDuration just returns that value? So then regardless when prometheus polls there's always a valid duration value to return for?

sunnywu:
could you also record in MR description what testing you have done so reviewers know the level of testing or give more suggestions if necessary

sunnywu:
how are you planning to setup the alert on "1+ successful job execution over last ~25 mins"? do we need to create a COunter and that get incremented upon every successful job execution and then we can setup the alert based on the rate of the counter increase (over last 25 mins)?

src/main/java/com/uid2/admin/job/JobDispatcher.java
aulme: Prom will gather those values periodically, by default I think it's 15sec but we could increase it if we prefer to improve resolution.
A gauge should gather the current value. If our metric is how long the current job has been running and the current job is null, that means currently the job is not running so it indeed has a running time of 0.
Alerting should be fine as long as duration we alert on is greater than scraping period. E.g. if we want to alert on jobs running >1min this should not be a problem.

sunnywu: I think 99% of the time there is no job being run when Prom polls so the gauge will always get 0 which is not informative? and we aren't synchronising `synchronized (jobLock)` inside getExecutionDuration which could create concurrency issue?

So if we use the gauge purely to detect hanged jobs then this is indeed ok tho i thought the gauge can be used as a metric for dashboard too (to see how the duration changes over time, e.g. months as more sites and operators get created). So getting 0 all the time for the gauge isn't useful and probably we need a separate gauge to check for lastSuccessfulRunDuration for us to plot and alert on increase in exec duration?

gmsdelmundo-ttd: Assigning currentJob to a temp variable shouldn't cause any concurrency issue, especially since the function only uses a getter function. However, I've changed it back to a synchronized lock

gmsdelmundo-ttd: And execution duration = 0 can just be interpreted as something is not running. If needed, on the same graph visualising the execution durations, we could also query from the logs when a job actually started executing, so we can get points together with the executionDuration line chart

sunnywu: as long as the end result is we get an alert if there's no successful job run in last ~25 mins then it's all fine regardless how you implement it. 

so atm we run a job every 20 mins, each time takes around 7-10 seconds so our of an hour (3600 seconds), only 30 out of 3600 seconds the execution duration is not null, so unless Prom polls every second (which I doubt?) otherwise Prom will most likely get 0 most of the time. If it's just for detecting long running/hanged job then it's all good.

gmsdelmundo-ttd:
@sunnywu

> could you also record in MR description what testing you have done so reviewers know the level of testing or give more suggestions if necessary
Sure

how are you planning to setup the alert on "1+ successful job execution over last ~25 mins"? do we need to create a COunter and that get incremented upon every successful job execution and then we can setup the alert based on the rate of the counter increase (over last 25 mins)?
The code change is just to set up the execution duration metrics. The alert to be set up will check if duration > 25 min, and to avoid repeated alerts, maybe just set it to max 3 alerts per hour

PR #55, opened 2023-02-15
Upgraded Vert.x to 4.3.8

ghost:


src/main/java/com/uid2/admin/store/writer/EncryptionKeyStoreWriter.java
```diff
@@ -32,7 +33,7 @@ public void upload(Collection<EncryptionKey> data, JsonObject extraMeta) throws
             json.put("created", key.getCreated().getEpochSecond());
             json.put("activates", key.getActivates().getEpochSecond());
             json.put("expires", key.getExpires().getEpochSecond());
-            json.put("secret", key.getKeyBytes());
+            json.put("secret", Base64.getEncoder().encodeToString(key.getKeyBytes()));
```
thomasm-ttd, line 36: Who reads this data? I assume this will change what is written.

src/main/java/com/uid2/admin/vertx/AdminVerticle.java
```diff
@@ -25,149 +27,115 @@ public class AdminVerticle extends AbstractVerticle {
...
-                    final JsonObject userInfo = res.result();
-                    ctx.response().end(userInfo.getString("email"));
-                }
-            });
-        });
+
+        final OAuth2Auth oauth2Provider = (OAuth2Auth) authFactory.createAuthProvider(vertx);
+
+        final AuthenticationHandler authHandler = authFactory.createAuthHandler(vertx, router.route("/oauth2-callback"), oauth2Provider);
+        final RedirectToRootHandler redirectToRootHandler = new RedirectToRootHandler();
```
thomasm-ttd, line 67: And where has the /protected endpoint gone?

src/main/java/com/uid2/admin/vertx/AdminVerticle.java
```diff
@@ -25,149 +27,115 @@ public class AdminVerticle extends AbstractVerticle {
...
-                } else {
-                    handleEmailContactInfo(contact, rc);
-                }
-            }
-        });
+    private void respondWithRealUser(RoutingContext rc, OAuth2Auth oauth2Provider) {
+        oauth2Provider.userInfo(rc.user())
+                .onFailure(e -> {
+                    rc.session().destroy();
+                    rc.fail(e);
```
thomasm-ttd, line 103: This will return the full error, not just the cause as previously. This could leak information

conf/local-config.json
```diff
@@ -28,5 +28,8 @@
   "salt_snapshot_activates_in_seconds": 180,
   "salt_snapshot_expires_after_seconds": 180000,
   "is_auth_disabled": true,
-  "filter_key_over_cut_off_days": false
+  "oauth2_callback_url": "http://localhost:8089/oauth2-callback",
+  "github_client_key": "generate client key",
```
cYKatherine, line 32: Have we set those value in `k8s-deployment`?

gmsdelmundo-ttd: Yep, but I just added it here in case anyone wanted to test OAuth 2.0 themselves (had to test it as part of the upgrade). The fields won't be used as long as `is_auth_disabled` is set to `true`, and I've updated the README to explain how to test OAuth 2.0 locally

src/main/java/com/uid2/admin/store/writer/EncryptionKeyStoreWriter.java
gmsdelmundo-ttd: The problem here is that the Base64 values actually get incorrectly encoded when writing to the site-specific files

One example (global file above, site-specific file below):
RgyxkP4yP1gYhCINq7O9PxM6jX+etPqSQluZxjB1aG8=
RgyxkP4yP1gYhCINq7O9PxM6jX-etPqSQluZxjB1aG8

The `+` became a `-`, and this causes an error when trying to decode the base64 string

gmsdelmundo-ttd: This `upload()` function is called mainly when writing files to S3

src/main/java/com/uid2/admin/vertx/AdminVerticle.java
gmsdelmundo-ttd: I can add it back, is it being used anywhere? I don't see any `/protected` API calls in any of the repos

src/main/java/com/uid2/admin/store/writer/EncryptionKeyStoreWriter.java
thomasm-ttd: If that is the case, then the readers of the S3 files will be expecting something different. I think this is very risky.

src/main/java/com/uid2/admin/vertx/AdminVerticle.java
thomasm-ttd: ok

src/main/java/com/uid2/admin/vertx/AdminVerticle.java
gmsdelmundo-ttd: Sure I can add it back just in case (though I'm pretty sure the `/protected` endpoint is copied from the Vert.x routing examples...)

cYKatherine (approved):

src/main/java/com/uid2/admin/store/writer/EncryptionKeyStoreWriter.java
gmsdelmundo-ttd: Yeah I think this might actually be an existing issue and will appear for special characters like + or /

Whatever is stored in `keyBytes` array is correct. The only issue here is converting the byte array to the actual base64 string

gmsdelmundo-ttd: So adding the `encodeToString()` function above should rectify the aforementioned inconsistency

gmsdelmundo-ttd: as mentioned in uid2-operator, we actually need all the Base64 encodes for `JsonObject.put()` with `byte[]` data types because of the Vert.x 4 deprecation of their `JsonObject` implementation

Vert.x 3 encodes the byte array as a base64-encoded string
```
    public JsonObject put(String key, byte[] value) {
        Objects.requireNonNull(key);
        this.map.put(key, value == null ? null : Base64.getEncoder().encodeToString(value));
        return this;
    }
```

Vert.x 4 keeps the byte array as is
```
  public JsonObject put(String key, Object value) {
    Objects.requireNonNull(key);
    map.put(key, value);
    return this;
  }
```

thomasm-ttd (approved):

PR #57, opened 2023-02-16
Create site key on publisher creation

mcollins-ttd:


src/main/java/com/uid2/admin/secret/IEncryptionKeyManager.java
```diff
@@ -4,4 +4,9 @@
 
 public interface IEncryptionKeyManager {
     EncryptionKey addSiteKey(int siteId) throws Exception;
+
+    /**
+     * Creates a site key, if none exists. If created, the key is active immediately.
+     */
+    EncryptionKey ensureSiteKeyExists(int siteId) throws Exception;
```
jon8787, line 11: I've seen "ensure" used to mean "check or throw"
Maybe something like "getOrCreateSiteKey"?

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -414,10 +422,18 @@ private void handleClientRoles(RoutingContext rc) {
...
             // return client with new key
             rc.response().end(jsonWriter.writeValueAsString(c));
         } catch (Exception e) {
             rc.fail(500, e);
         }
     }
+
+    private void ensurePublisherSiteKeyExists(ClientKey clientKey) throws Exception {
+        if (clientKey.hasRole(Role.GENERATOR)) {
+            this.encryptionKeyManager.ensureSiteKeyExists(clientKey.getSiteId());
```
jon8787, line 436: is the return value ever used or could ensureSiteKeyExists return void?

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -414,10 +422,18 @@ private void handleClientRoles(RoutingContext rc) {
...
+
             // return client with new key
             rc.response().end(jsonWriter.writeValueAsString(c));
         } catch (Exception e) {
             rc.fail(500, e);
         }
     }
+
+    private void ensurePublisherSiteKeyExists(ClientKey clientKey) throws Exception {
+        if (clientKey.hasRole(Role.GENERATOR)) {
```
jon8787, line 435: actually Role.SHARER should also have site keys

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -414,10 +422,18 @@ private void handleClientRoles(RoutingContext rc) {
             // upload to storage
             storeWriter.upload(clients, null);
 
+            ensurePublisherSiteKeyExists(c);
```
jon8787, line 425: I wonder if there's any deduplication we could do here. eg add & update  both upload/ensure/respond, and future changes to these would happen in sync

src/main/java/com/uid2/admin/secret/IEncryptionKeyManager.java
mcollins-ttd: 👍 Renamed to `createSiteKeyIfNoneExists`

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
mcollins-ttd: Leaving for now, but let's come back to it when we have settled on the new roles.

jon8787 (approved):

PR #72, opened 2023-05-09
Cbc UI d2 1052 api endpoints to support portal

cody-constine-ttd:


src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -0,0 +1,187 @@
...
+        this.auth = auth;
+        this.writeLock = writeLock;
+        this.storeWriter = storeWriter;
+        this.keyAclProvider = keyAclProvider;
+        this.siteProvider = siteProvider;
+    }
+
+    @Override
+    public void setupRoutes(Router router) {
+        router.get("/api/sharing/list/get").handler(
```
lionell-pack-ttd, line 44: Are we using the term 'list' for anything? In an API URL I'd normally expect names and plurals - so maybe:
/api/sharing/sites
/api/sharing/sites/:siteid
That kind of thing.

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -0,0 +1,187 @@
...
+                auth.handle(this::handleKeyAclListAll, Role.SHARING_PORTAL)
+        );
+        router.get("/api/sharing/list/:siteId/get").handler(
```
lionell-pack-ttd, line 47: I don't think the `/get` and `/set` parts of the URL are needed - it's commonly understood that if you do a GET request to a resource, you'll get that resource, and a POST (or PUT) will be setting it.

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -0,0 +1,187 @@
...
+            return;
+        }
+
+        final EncryptionKeyAcl newAcl = new EncryptionKeyAcl(true, whitelist.stream()
+                .map(s -> (Integer) s)
+                .collect(Collectors.toSet()));
+
+        collection.put(site_id, newAcl);
+        try {
+            storeWriter.upload(collection, null);
```
lionell-pack-ttd, line 167: Do we need some locking from the moment we read until when we write back? Multiple requests could come in at around the same time. I see there's a writeLock injected but it doesn't seem to be used.

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -0,0 +1,187 @@
...
+
+        rc.response()
+                .putHeader(HttpHeaders.CONTENT_TYPE, "application/json")
+                .end(jo.encode());
+
+    }
+
+    private int computeWhitelistHash(Set<Integer> list)
+    {
+        return Objects.hash(list);
```
mcollins-ttd, line 185: It's relatively easy to get a collision here:

> The hash code of a map is defined to be the sum of the hash codes of each entry in the map's entrySet() view.

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/AbstractMap.html#hashCode()

```
Objects.hash(new HashSet<>(Arrays.asList(5, 7))) == Objects.hash(new HashSet<>(Arrays.asList(2, 10)))
```

src/main/java/com/uid2/admin/vertx/service/SharingService.java
cody-constine-ttd: I was thinking lists because we are getting/setting lists not sites

lionell-pack-ttd: Yeah, let's run with it then :)

PR #73, opened 2023-05-15
UID2-1113 Do not create .bak files when writing files in S3

thomasm-ttd:
Update the tags for salts to either current or obsolete


src/main/java/com/uid2/admin/store/writer/SaltStoreWriter.java
```diff
@@ -106,6 +127,22 @@ private void uploadSaltsSnapshot(RotatingSaltProvider.SaltSnapshot snapshot, Str
...
+    private void setStatusTagToCurrent(String location) throws CloudStorageException {
+        this.cloudStorage.setTags(location, this.getCurrentTags());
+    }
+
+    private void setStatusTagToObsolete(String location) throws CloudStorageException {
+        this.cloudStorage.setTags(location, getObsoleteTags());
+    }
+
+    private Map<String, String> getCurrentTags(){
+        return Map.of("status", "current");
```
vishalegbert-ttd, line 142: Can make this constant

thomasm-ttd: Good idea, Done

PR #75, opened 2023-06-01
UID2-860 Add rename key functionality to admin panel

cYKatherine:
image

The above will be calling the backend with the below payloads:
image


src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -431,6 +437,39 @@ private void handleClientRoles(RoutingContext rc) {
         }
     }
 
+    private void handleClientRename(RoutingContext rc) {
```
vishalegbert-ttd, line 440: Can you add some tests to ClientKeyServiceTest? Should check if it works and should also check if it prevents a user from renaming to a non-unique name

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -431,6 +437,39 @@ private void handleClientRoles(RoutingContext rc) {
...
+            existingClient.withNameAndContact(newName);
+
+            List<ClientKey> clients = this.clientKeyProvider.getAll()
+                    .stream().sorted((a, b) -> (int) (a.getCreated() - b.getCreated()))
+                    .collect(Collectors.toList());
+
+            // upload to storage
+            storeWriter.upload(clients, null);
+
+            createPublisherSiteKeyIfNoneExists(existingClient);
```
vishalegbert-ttd, line 464: Do we need this?

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
cYKatherine: Done

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
cYKatherine: Removed

vishalegbert-ttd (approved):

PR #85, opened 2023-06-26
Cbc UI d2 628 adding keysets

cody-constine-ttd:


pom.xml
```diff
@@ -16,7 +16,7 @@
         <!-- check micrometer.version vertx-micrometer-metrics consumes before bumping up -->
         <micrometer.version>1.1.0</micrometer.version>
         <junit-jupiter.version>5.7.0</junit-jupiter.version>
-        <uid2-shared.version>2.5.0-6429c70acc</uid2-shared.version>
+        <uid2-shared.version>2.5.6-SNAPSHOT</uid2-shared.version>
```
aulme, line 19: Should be reverted for the commit

cody-constine-ttd: Yeah it will need to be changed to the real release when it happens

src/main/java/com/uid2/admin/job/jobsync/PrivateSiteDataSyncJob.java
```diff
@@ -4,27 +4,24 @@
 import com.uid2.admin.job.jobsync.acl.KeyAclSyncJob;
 import com.uid2.admin.job.jobsync.client.ClientKeySyncJob;
 import com.uid2.admin.job.jobsync.key.EncryptionKeySyncJob;
+import com.uid2.admin.job.jobsync.key.KeysetKeySyncJob;
```
jon8787, line 7: hello

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -283,13 +315,39 @@ private void handleRotateSiteKey(RoutingContext rc) {
...
+
     private void handleRotateAllSiteKeys(RoutingContext rc) {
         try {
-            final RotationResult result = rotateKeys(rc, siteKeyActivatesIn, siteKeyExpiresAfter, s -> SiteUtil.isValidSiteId(s) || s == Const.Data.AdvertisingTokenSiteId);
+            final RotationResult<EncryptionKey> result = rotateKeys(rc, siteKeyActivatesIn, siteKeyExpiresAfter, s -> SiteUtil.isValidSiteId(s) || s == Const.Data.AdvertisingTokenSiteId);
             if (result == null) {
                 return;
             }
 
+            //rotate all keyset keys when site keys rotate
```
Ian-Nara, line 350: does not look like the keyset keys are being rotated here?

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -15,82 +15,227 @@
...
+               this.keyManager.addKeysetKey(keysetId);
+           } catch (Exception e) {
+               rc.fail(500, e);
+               return;
+           }
+
+
+           JsonObject jo = new JsonObject();
+           jo.put("site_id", newKeyset.getSiteId());
+           jo.put("keyset_id", newKeyset.getKeysetId());
```
jon8787, line 114: should keyset_id go first ?

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -15,82 +15,227 @@
...
+            LOGGER.warn("Failed to find keyset for keyset id: " + keyset);
+            rc.fail(404);
+            return;
+        }
+
+        JsonArray listedSites = new JsonArray();
+        keyset.getAllowedSites().stream().sorted().forEach((listedSiteId) -> listedSites.add(listedSiteId));
+        JsonObject jo = new JsonObject();
+        jo.put("site_id", keyset.getSiteId());
+        jo.put("keyset_id", keyset.getKeysetId());
```
jon8787, line 149: this code looks duplicated from above, can we use a function?

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -15,82 +15,227 @@
...
-            site_id = Integer.parseInt(rc.pathParam("siteId"));
+            JsonArray ja = new JsonArray();
+            Map<Integer, Keyset> collection = this.keysetProvider.getSnapshot().getAllKeysets();
+            for (Map.Entry<Integer, Keyset> keyset : collection.entrySet()) {
+                JsonArray listedSites = new JsonArray();
+                keyset.getValue().getAllowedSites().stream().sorted().forEach((listedSiteId) -> listedSites.add(listedSiteId));
+
+                JsonObject jo = new JsonObject();
+                jo.put("site_id", keyset.getValue().getSiteId());
+                jo.put("keyset_id", keyset.getValue().getKeysetId());
```
jon8787, line 180: this code looks duplicated from above, can we use a function?

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -15,82 +15,227 @@
...
+        jo.put("allowlist", listedSites);
+        jo.put("hash", keyset.hashCode());
 
         rc.response()
                 .putHeader(HttpHeaders.CONTENT_TYPE, "application/json")
                 .end(jo.encode());
     }
 
-    private void handleKeyAclListAll(RoutingContext rc) {
+    private void handleAllowlistListALL(RoutingContext rc) {
```
jon8787, line 226: why the caps :P

src/main/resources/localstack/s3/admins/admins.json
```diff
@@ -8,7 +8,8 @@
       "ADMINISTRATOR",
       "CLIENTKEY_ISSUER",
       "SECRET_MANAGER",
-      "OPERATOR_MANAGER"
+      "OPERATOR_MANAGER",
+      "SHARING_PORTAL"
```
jon8787, line 12: is this for a different ticket?

src/main/java/com/uid2/admin/vertx/service/SharingService.java
aulme: Json is not ordered by spec so should be irrelevant.

On another bit, JsonObject can be more readably defined through:
```
JsonObject jo = JsonObject.of(
                    "name", newKeyset.getName(),
                    "allowlist", newKeyset.getAllowedSites(),
                     ...
            )
```

Would make it more readable than a lot of puts.

Also perhaps can give it a better name than `jo` - eg `responseBody`

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -15,82 +15,227 @@
...
+               rc.fail(500);
+           }
+
+            final JsonObject body = rc.body().asJsonObject();
+
+           final JsonArray whitelist = body.getJsonArray("allowlist");
+           Integer keysetId = body.getInteger("keyset_id");
+           final Integer siteId = body.getInteger("site_id");
+
+           final Map<Integer, Keyset> collection = this.keysetProvider.getSnapshot().getAllKeysets();
```
aulme, line 81: The name is very generic and even slightly inaccurate. How about `keysetsById` or something similar?

src/main/java/com/uid2/admin/vertx/service/SharingService.java
jon8787: Might be irrelevant for computers, but not for humans. Eg we wouldn't want every keyset to have elements in a different order

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -15,82 +15,227 @@
...
-                auth.handle(this::handleKeyAclListAll, Role.SHARING_PORTAL)
+                auth.handle(this::handleAllowlistListALL, Role.SHARING_PORTAL)
         );
         router.get("/api/sharing/list/:siteId").handler(
-                auth.handle(this::handleKeyAclList, Role.SHARING_PORTAL)
+                auth.handle(this::handleAllowlistList, Role.SHARING_PORTAL)
         );
         router.post("/api/sharing/list/:siteId").handler(
-                auth.handle(this::handleKeyAclSet, Role.SHARING_PORTAL)
+                auth.handle(this::handleAllowlistSet, Role.SHARING_PORTAL)
```
aulme, line 52: Good stuff that we do semantic POST/GET convention!
Was quite confused on the handler methods though - `handleAllowlistList` and `handleAllowlistSet` sound like they'll return allowlists as a list or as a set. Perhaps we could rename them to something like `handleSetAllowlist` or even better - `setAllowlist`?

aulme: Also is it still reasonable to have those consistently be `GET /api/sharing/lists`, `GET /api/sharing/lists/:siteId` and `POST /api/sharing/lists/:siteId`? Now they're not consistent with standard HTTP REST conventions.

src/main/java/com/uid2/admin/vertx/service/SharingService.java
aulme: I mean even our JSON library is not required to preserve the order in which we've added those properties.

jon8787: How do we have a consistent and sensible order in the current Admin GUI "List Keys"?

aulme: It doesn't randomize them, just doesn't provide guarantees on order being preserved

aulme: But yeah - as long as we don't actually rely on that, it's fine

src/main/java/com/uid2/admin/vertx/service/SharingService.java
cody-constine-ttd: Also Handle is kind of our convention from the other API, so I am going to preserve it here.

jon8787 (approved):

PR #87, opened 2023-07-05
Ian UI d2 627 admin keysets UI

Ian-Nara:


webroot/adm/default-keysets.html
```diff
@@ -0,0 +1,92 @@
...
+<h1>UID2 Admin - Default Keyset Access Management (Sharer)</h1>
+
+<a href="/">Back</a>
+
+<br>
+<br>
+
+<h3>Inputs</h3>
+
+<label for="siteId">Site Id:</label>
```
cody-constine-ttd, line 16: Should this be keyset id?

webroot/adm/default-keysets.html
```diff
@@ -0,0 +1,92 @@
...
+
+<div id="output">
+  <pre id="errorOutput"></pre>
+  <pre id="standardOutput"></pre>
+</div>
+
+<script language="JavaScript">
+  $(document).ready(function () {
+    $('#doListAll').on('click', function () {
+      doApiCall('GET', '/api/sharing/lists', '#standardOutput', '#errorOutput');
```
cody-constine-ttd, line 47: can you use the /api/sharing/keyset endpoints?

webroot/adm/keysets.html
```diff
@@ -0,0 +1,103 @@
...
+<h1>UID2 Admin - Keyset Access Management (Administrator)</h1>
+
+<a href="/">Back</a>
+
+<br>
+<br>
+
+<h3>Inputs</h3>
+
+<label for="siteId">Site Id:</label>
```
cody-constine-ttd, line 16: Same as above

webroot/adm/default-keysets.html
Ian-Nara: have keysets.html for the keyset endpoints and default-keysets.html for the list endpoints. Since they have almost the same functionality but different permissions I think we need to have different pages for them. If we don't need a UI for the list endpoints though I can get rid of that page

webroot/adm/keysets.html
Ian-Nara: Need to have both siteId and keysetId. The endpoint /api/sharing/keyset is consuming the site id from the request payload.

webroot/adm/default-keysets.html
Ian-Nara: /list endpoints are using siteId

webroot/adm/keysets.html
```diff
@@ -0,0 +1,103 @@
...
+<input type="text" id="siteId" name="siteId">
+<label for="keysetId">Keyset Id:</label>
+<input type="text" id="keysetId" name="siteId">
```
cody-constine-ttd, line 19: Copy paste error on the name

cody-constine-ttd (approved):

PR #93, opened 2023-07-18
Using IAM authentication flow for admin

aulme:


pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-admin</artifactId>
-    <version>2.10.0-7a209d003f</version>
+    <version>2.10.4-a38de4d955</version>
```
gmsdelmundo, line 9: Why manually update the version here? I thought our GH action will do that for us

src/main/java/com/uid2/admin/Main.java
```diff
@@ -206,6 +206,14 @@ public void run() {
         }
     }
 
+    private TaggableCloudStorage createStorage() {
```
gmsdelmundo, line 209: Should this be a util function instead? Since it's used here and in PrivateSiteDataSyncJob

pom.xml
aulme: That's from me doing a branch build before

src/main/java/com/uid2/admin/Main.java
aulme: It's temporary for backwards compatibility. The idea is to delete it once we're back to one auth method.

PR #96, opened 2023-07-21
improve keysets ui

Ian-Nara:
Changes in Keyset management:
- Keyset can be named
- can list keysets by site id
Changes in Encryption Key Management:
- can list keyset keys, list keys, list keyset keys by keyset id, list keys by site id


cody-constine-ttd (approved):
LGTM, tested all changes locally and looks to work better

src/main/resources/localstack/s3/keyset_keys/keyset_keys.json
```diff
@@ -1 +1,51 @@
...
+    "secret" : "bVZZOwCuO0fUf2P6CrwaY23erMErVQiMkGgpy2OMSxc=",
+    "keyset_id": 4
+  },
+  {
+    "id" : 5,
+    "created" : 1609459200,
+    "activates" : 1609459200,
+    "expires" : 1893456000,
+    "secret" : "zpxOafFdffTpeYFvaWnJlaLzJNzk+6qi77U2Umh7QJ4=",
+    "keyset_id": 5
```
sunnywu, line 40: the keysets.json under localstack shows there's no keyset with id 5 and 6 like below so this caused NullPointerException problem in com.uid2.admin.util.PrivateSiteUtil#getKeysetKeys when

globalKeysets.get(keysetKey.getKeysetId()) tries to globalKeysets.get an non-existing keysetId.

Easiest is to reassign keyset_id above to 821 and the below to something else

sunnywu:
Could you help improve the code on com.uid2.admin.util.PrivateSiteUtil#getKeysetKeys and don't crash if 

globalKeysets.get(keysetKey.getKeysetId()) is not found?

webroot/adm/keysets.html
```diff
@@ -69,6 +91,7 @@ <h3>Output</h3>
     $('#doUpdate').on('click', function () {
       const siteId = parseInt($('#siteId').val());
```
sunnywu, line 92: maybe out of scope for this ticket, just noticed this when i'm testing - what is the purpose of "Update Keyset"?  
so it can do 

1. update the Site Id this keyset is for
2. Update the keyset name
3 Add/remove allow site ids

There are a few issues:

1. Turns out Name/Add/Remove SiteId fields are optional but Site Id is compulsory even if we are not changing it. Is this really necessary if Keyset id is always required and we aren't changing site id? 
2. If Keyset Id is missing, it just said Error: 400: Bad Request. Could we improve it so it said Keyset Id is missing and that's a must have?
3. there is no check on the validity of site ids on Site Id (i can put random number into associated Site Id and the update will be successful), Add SiteIds and Remove Site Ids. Should there be otherwise people can mistypes and mistakes are ignored?

I can't remember if site id values arechecked before keyset work.


@jon8787 @cody-constine-ttd @zaiweidu thoughts?

Ian-Nara: 1.  So "update keyset" uses the same endpoint as "make new keyset", with the siteid being used to create a new keyset if no keyset id is provided. But if keyset id is provided then site id is not needed. Should we even allow an existing keyset to have it's site id reassigned to a different site? Versus just make new keyset for the other site id?
2. Putting in a check on the ui side for if keysetId/siteId is not there, as otherwise it ends up trying to call it with "NaN". I also updated the backend to return more helpful message. 
3. I can't see any reason not to do this. I have added this check in and updated tests in SharingServiceTest.

webroot/adm/keysets.html
```diff
@@ -28,10 +30,11 @@ <h3>Inputs</h3>
 <h3>Operations</h3>
 
 <ul>
-  <li class="ro-adm" style="display: none"><a href="#" id="doListAll">List All</a></li>
-  <li class="ro-adm" style="display: none"><a href="#" id="doList">List</a></li>
-  <li class="ro-adm" style="display: none"><a href="#" id="doReset">Reset</a></li>
-  <li class="ro-adm" style="display: none"><a href="#" id="doUpdate">Update</a></li>
+  <li class="ro-adm" style="display: none"><a href="#" id="doListAll">List All Keysets</a></li>
+  <li class="ro-adm" style="display: none"><a href="#" id="doListAllBySite">List All Keysets By Site Id</a></li>
+  <li class="ro-adm" style="display: none"><a href="#" id="doListKeyset">List Keyset</a></li>
```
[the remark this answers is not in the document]
Ian-Nara: changed text

src/main/resources/localstack/s3/keyset_keys/keyset_keys.json
Ian-Nara: I have tested locally. Had keysets 1-6 before, I must have overwritten this at some point. I have updated the localstack now

Ian-Nara:
> For the issue i mentioned above, could you help improve the code on com.uid2.admin.util.PrivateSiteUtil#getKeysetKeys and don't crash if
> 
> globalKeysets.get(keysetKey.getKeysetId()) is not found?
> 
> And use the logger to log an error if it is indeed not found.
> 
> Please look at existing tests and see if you can add one for this as well



> For the issue i mentioned above, could you help improve the code on com.uid2.admin.util.PrivateSiteUtil#getKeysetKeys and don't crash if
> 
> globalKeysets.get(keysetKey.getKeysetId()) is not found?
> 
> And use the logger to log an error if it is indeed not found.
> 
> Please look at existing tests and see if you can add one for this as well

I have put a check for this in PrivateSiteUtil and have added a KeysetKeys test section with a couple tests including for this scenario.

webroot/adm/keysets.html
sunnywu: For 1., will let you update the code from slack chat result first.
For 2., the update looks ok
For 3, is there a check on _Add Site id_'s value? I _allowed_ site id 33 as shown in screencap which doesn't exist but it still allows me adding it and shown in the output? 
image

And currently if the Site ID field is set to invalid site id and i click update, there's only a error logged in admin service log like:
_ERROR com.uid2.admin.vertx.ResponseUtil - Site id 999 not valid_

but nothing shown on the html web page Output section? Probably wanna check when the API call return errors, whether those errors will appear on the webpage output

src/main/resources/localstack/s3/keyset_keys/keyset_keys.json
sunnywu: similar to my other comment i would suggest to not use same values for keyset id and site id in the localstack json files. and i think we need some more realistic allow_lists values as currently all of the site ids in all allow_lists are invalid site ids (e.g. 2,3,4)?

webroot/adm/keysets.html
zaiweidu: @Ian-Nara for #1, we do NOT allow Keyset's siteId change once it's created.

src/test/java/com/uid2/admin/util/PrivateSiteUtilTest.java
```diff
@@ -589,6 +595,169 @@ public void testGenerateEncryptionKeyAclDataForEachSite() {
...
     }
 
+    @Nested
+    class KeysetKeys {
+
+        KeysetKey keysetKey1 = new KeysetKey(1, new byte[]{}, Instant.now(), Instant.now(), Instant.now(), 1);
+        KeysetKey keysetKey2 = new KeysetKey(2, new byte[]{}, Instant.now(), Instant.now(), Instant.now(), 2);
+        KeysetKey keysetKey3 = new KeysetKey(3, new byte[]{}, Instant.now(), Instant.now(), Instant.now(), 3);
+
+        Keyset keyset1 = new Keyset(1, 1, "test-name-1", ImmutableSet.of(), 999999, true, true);
```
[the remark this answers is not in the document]
Ian-Nara: I have updated the Ids in PrivateSiteUtilTest.java

webroot/adm/keysets.html
Ian-Nara: Right now it just checks the keyset owner's site to be valid. I will add allowlist also, but am moving this to another branch

Ian-Nara: Also fixed error output not showing for update/reset

src/main/resources/localstack/s3/keyset_keys/keyset_keys.json
Ian-Nara: updated to make the lists more realistic and ids more distinctive

webroot/adm/keysets.html
```diff
@@ -78,22 +110,44 @@ <h3>Output</h3>
...
+        $('#errorOutput').html("Required Parameters: site_id")
+        return;
+      }
+      let name = $('#keysetName').val();
       const url = '/api/sharing/keyset';
-      doApiCall('POST', url, '#standardOutput', 'errorOutput', JSON.stringify({allowlist: [] ,site_id: siteId}));
+
+      let payload;
+      if(name && (name = $.trim(name))) {
+        payload = JSON.stringify({allowlist: [], site_id: siteId, name: name})
```
sunnywu, line 145: @jon8787 @zaiweidu when we create new keyset, should we allow people to provide a allowlist straight away instead of making adding allowlist as a separate second request?

webroot/adm/keysets.html
```diff
@@ -17,6 +17,8 @@ <h3>Inputs</h3>
 <input type="text" id="siteId" name="siteId">
 <label for="keysetId">Keyset Id:</label>
 <input type="text" id="keysetId" name="keysetId">
+<label for="keysetName">Set Keyset Name:</label>
+<input type="text" id="keysetName" name="keysetName">
 <label for="addSiteIds">Add Site Ids:</label>
```
sunnywu, line 22: I still think _Add/Remove Site Ids_ should be renamed to something more descriptive like 
_Add Site Ids to Allowed Sites_ and _Remove Site Ids from Allowed Sites_ 

to be more descriptive as Add Site Ids could be confusing to the _Site Id_ field

webroot/adm/keysets.html
```diff
@@ -17,6 +17,8 @@ <h3>Inputs</h3>
 <input type="text" id="siteId" name="siteId">
 <label for="keysetId">Keyset Id:</label>
```
jon8787, line 18: can we make keyset id appear before site id?

webroot/adm/keysets.html
jon8787: I like the idea of "Set" but I'm worried about people blowing away a large list accidentally. How would that be solved?

webroot/adm/keysets.html
jon8787: Re "allowed_sites straight away", let's solve this later when we offer multiple keysets. For now, people won't generally be adding keysets since the first one is created automatically.

jon8787 (approved):

zaiweidu (changes requested):

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -65,51 +71,62 @@ public void setupRoutes(Router router) {
...
-               keysetProvider.loadContent();
-           } catch (Exception e) {
-               LOGGER.error("Failed to load key acls");
-               rc.fail(500);
-           }
+            try {
+                keysetProvider.loadContent();
+                siteProvider.loadContent();
+            } catch (Exception e) {
+                LOGGER.error("Failed to load key acls");
```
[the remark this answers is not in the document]
Ian-Nara: fixed

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -237,71 +253,71 @@ private void handleListAllAllowlist(RoutingContext rc) {
...
+            } catch (Exception e) {
+                LOGGER.warn("Failed to parse a site id from list request", e);
+                rc.fail(400, e);
+                return;
+            }
+
+            try {
+                keysetProvider.loadContent();
+            } catch (Exception e) {
+                LOGGER.error("Failed to load key acls");
```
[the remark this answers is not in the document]
Ian-Nara: fixed

zaiweidu (approved):

PR #97, opened 2023-07-24
added logic for the keysets to be created automatically

cody-constine-ttd:


src/main/java/com/uid2/admin/util/KeysetUtil.java
```diff
@@ -20,7 +23,8 @@ public static AdminKeyset lookUpKeyset(int siteId, Map<Integer, AdminKeyset> key
     }
 
     public static Integer getMaxKeyset(Map<Integer, AdminKeyset> keysets) {
-        return keysets.keySet().size();
+        if(keysets.isEmpty()) return 3;
```
jon8787, line 26: Can we avoid calling getMaxKeyset() until after we've added the initial 3 keysets?

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -551,6 +551,15 @@ private int getOrCreateKeysetId(int siteId)
         AdminKeyset keyset = lookUpKeyset(siteId, currentKeysets);
         if(keyset == null) {
             int newKeysetId = getMaxKeyset(currentKeysets)+1;
+            if(siteId == Const.Data.MasterKeySiteId) {
```
jon8787, line 554: where are the calls to getOrCreateKeysetId() that will create these special keysets?

src/main/java/com/uid2/admin/util/KeysetUtil.java
sunnywu: Worst case of having non incrementing keyset ids should not cause any problems.... probably good idea to verify and log error if this is so. but in integ/prod there must be master/refresh/pub key ids so shouldn't be a problem

jon8787: Should we have test coverage? Ideally I'd like to see a test that fails if we returned 2 from here instead of 3.

sunnywu:
closing this PR as I have rebase off master instead in [this ](https://github.com/IABTechLab/uid2-admin/pull/98)PR

PR #98, opened 2023-07-27
Syw UId2-1534 reserve right keyset id for special encryption keys

sunnywu:
1.  Change 'Add Site Ids' to 'Add Allowed Site Ids' and 'Remove Allowed Site Ids' to be less confusing
2. renamed all allowlists to allowed_sites
3.  Port Cody's changes from https://github.com/IABTechLab/uid2-admin/pull/97 to rebase off master
4. reserve right keyset id for special encryption keys

### DIFF
--- a/src/main/java/com/uid2/admin/util/KeysetUtil.java
+++ b/src/main/java/com/uid2/admin/util/KeysetUtil.java
@@ -3,9 +3,12 @@ package com.uid2.admin.util;
 import com.uid2.shared.auth.Keyset;
 
 import java.time.Instant;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.stream.Collectors;
+
+import static java.lang.Math.max;
 
 public class KeysetUtil {
     public static Keyset lookUpKeyset(int siteId, Map<Integer, Keyset> keysets) {
@@ -18,7 +21,8 @@ public class KeysetUtil {
     }
 
     public static Integer getMaxKeyset(Map<Integer, Keyset> keysets) {
-        return keysets.keySet().size();
+        if(keysets.isEmpty()) return 3;
+        return max(Collections.max(keysets.keySet()), 3);
     }
 
     public static Keyset createDefaultKeyset(int siteId, int keysetId) {
--- a/src/main/java/com/uid2/admin/util/KeysetUtil.java
+++ b/src/main/java/com/uid2/admin/util/KeysetUtil.java
@@ -21,6 +21,8 @@ public class KeysetUtil {
     }
 
     public static Integer getMaxKeyset(Map<Integer, Keyset> keysets) {
+        // keyset id 1/2/3 are assigned for master/refresh/default publisher encryption key ids,
+        // so we always reserve these 3 keyset ids for them
         if(keysets.isEmpty()) return 3;
         return max(Collections.max(keysets.keySet()), 3);
     }
--- a/src/main/java/com/uid2/admin/util/KeysetUtil.java
+++ b/src/main/java/com/uid2/admin/util/KeysetUtil.java
@@ -12,6 +12,12 @@ import java.util.stream.Collectors;
 import static java.lang.Math.max;
 
 public class KeysetUtil {
+
+    public static final String MasterKeysetName = "Master";
+    public static final String RefreshKeysetName = "Refresh";
+    public static final String FallbackPublisherKeysetName = "Fallback Publisher";
+
+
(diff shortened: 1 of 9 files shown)


src/main/java/com/uid2/admin/util/KeysetUtil.java
```diff
@@ -18,7 +21,10 @@ public static Keyset lookUpKeyset(int siteId, Map<Integer, Keyset> keysets) {
     }
 
     public static Integer getMaxKeyset(Map<Integer, Keyset> keysets) {
-        return keysets.keySet().size();
+        // keyset id 1/2/3 are assigned for master/refresh/default publisher encryption key ids,
+        // so we always reserve these 3 keyset ids for them
+        if(keysets.isEmpty()) return 3;
+        return max(Collections.max(keysets.keySet()), 3);
```
zaiweidu, line 27: can we use the consts defined in _com.uid2.shared.Const.Data_ instead of 1/2/3?

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -548,6 +548,15 @@ private int getOrCreateKeysetId(int siteId)
         Keyset keyset = lookUpKeyset(siteId, currentKeysets);
         if(keyset == null) {
             int newKeysetId = getMaxKeyset(currentKeysets)+1;
+            if(siteId == Const.Data.MasterKeySiteId) {
+                newKeysetId = Const.Data.MasterKeysetId;
+            }
+            if(siteId == Const.Data.RefreshKeySiteId) {
```
zaiweidu, line 554: 'else-if' is better here as only 1 of 3 cases will match. no need to compare 3 times.

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -120,7 +120,7 @@ private JsonObject jsonFullKeyset(Keyset keyset) {
         jo.put("keyset_id", keyset.getKeysetId());
         jo.put("site_id", keyset.getSiteId());
         jo.put("name", keyset.getName());
-        jo.put("allowlist", keyset.getAllowedSites());
+        jo.put("allowed_sites", keyset.getAllowedSites());
```
zaiweidu, line 123: we allow ID_READER to access all keysets whose 'allowed_site' equals to 'null'. I wonder when you're going to set it to null in Admin.

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
```diff
@@ -548,6 +548,15 @@ private int getOrCreateKeysetId(int siteId)
         Keyset keyset = lookUpKeyset(siteId, currentKeysets);
         if(keyset == null) {
             int newKeysetId = getMaxKeyset(currentKeysets)+1;
+            if(siteId == Const.Data.MasterKeySiteId) {
+                newKeysetId = Const.Data.MasterKeysetId;
```
jon8787, line 552: just had a thought - [Migration stage -> keyset.json](https://atlassian.thetradedesk.com/confluence/display/UID2/UID2-545+Sharing+-+Data+Storage+for+Keys+and+Grants#UID2545SharingDataStorageforKeysandGrants-MigrationStage-MultipleKeysets) shows that name should be populated for the special keysets, eg "Master", "Refresh".

src/main/java/com/uid2/admin/vertx/service/SharingService.java
jon8787: Keep in mind this is old code, only the field name is being changed. Looking at https://admin-integ.uidapi.com/adm/keysets.html , most allowlists are null.

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
sunnywu: fixed

src/main/java/com/uid2/admin/vertx/service/EncryptionKeyService.java
sunnywu: Added both and  "FallbackPublisher" keyset as well

sunnywu: the name are defined in this shared lib PR https://github.com/IABTechLab/uid2-shared/pull/99/files

sunnywu:
Tested manually by removing the existing keyset/keysetkeys files in localstack and on admin service startup, the name of the speical master/refresh/failedpublisher keysetname are all set correctly.

PR #100, opened 2023-07-28
add gcp enclave page V2.

yishi-ttd:
Add a new page for GCP enclave v2 (OIDC).
 - Mimiced existing GCP enclave page.

Screenshot:
image

### DIFF
--- a/webroot/adm/enclave-gcp-v2.html
+++ b/webroot/adm/enclave-gcp-v2.html
@@ -37,15 +37,6 @@
 <br>
 <br>
 
-<label for="identityScope" class="bold">Identity Scope:</label>
-<br>
-<select id="identityScope">
-  <option value="uid2">UID2</option>
-  <option value="euid">EUID</option>
-</select>
-<br>
-<br>
-
 <label for="debugMode" class="bold">DebugMode:</label>
 <br>
 <label for="debugMode">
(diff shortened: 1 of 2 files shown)


webroot/adm/enclave-gcp-v2.html
```diff
@@ -0,0 +1,191 @@
...
+      console.log("enclaveId:" + enclaveId);
+      $('#enclaveId').text(enclaveId);
+    });
+
+    // update gcloud Cli
+    const imageFamily = isDebug ? 'confidential-space-debug' : 'confidential-space';
+    const cli = gcloudCli
+            .replaceAll('<IMAGE_FAMILY>', imageFamily)
+            .replaceAll('<IMAGE_SHA>', imageSha)
+            .replaceAll('<API_TOKEN>', apiToken)
```
thomasm-ttd, line 166: The API token is considered sensitive and we never send it via email, or in any text files. So the resulting gcloud CLI command must still have <API_TOKEN> (the actual string, not a replaced value). As the API key isn't used for the generation of the enclave Id, there is no reason for it to be collected on this page.

webroot/adm/enclave-gcp-v2.html
```diff
@@ -0,0 +1,191 @@
...
+<h1>UID2 Admin - GCP Enclave ID Tool(V2)</h1>
+
+<a href="/">Back</a>
+
+<br>
+<br>
+
+<h3>Inputs</h3>
+
+<label for="apiToken" class="bold">Operator Api Key:</label>
```
thomasm-ttd, line 23: We should not be collecting the API key here - we don't need it for the enclave Id, so we should not collect it. It is considered to be a secret.

yishi-ttd: ACK, will update.

webroot/adm/enclave-gcp-v2.html
yishi-ttd: ACK, will update.

webroot/adm/enclave-gcp-v2.html
```diff
@@ -0,0 +1,173 @@
...
+<br>
+
+<label for="enclaveId">Generated GCP Enclave ID:</label>
+<br>
+<pre><label id="enclaveId">TBD</label></pre>
+<br>
+<br>
+
+<script language="JavaScript">
+  const gcloudCli = '$ gcloud compute instances create <INSTANCE_NAME> \\\n\
```
lunwang-ttd, line 86: Should we add a placeholder for zone config so that partners can change it?

lunwang-ttd: Also for `machine-type`

yishi-ttd: There would be too many GCP overrides. I would suggest to only provide necessary commands for partners. i.e. they will use default zone & machine-type if not specified, but OK to change. We could mention this in public doc guide.
WDYT?

thomasm-ttd: I agree, we can document what else they can change

thomasm-ttd (approved):

webroot/adm/enclave-gcp-v2.html
lunwang-ttd: I see, a well written doc will be good enough

lunwang-ttd (approved):

PR #106, opened 2023-08-02
Cbc UI d2 1570 admin feature switch

cody-constine-ttd:


src/main/java/com/uid2/admin/store/writer/KeysetKeyStoreWriter.java
```diff
@@ -9,21 +9,31 @@
...
         FileName dataFile = new FileName("keyset_keys", ".json");
         String dataType = "keyset_keys";
         writer = new ScopedStoreWriter(provider, fileManager, versionGenerator, clock, scope, dataFile, dataType);
+        this.enableKeysets = enableKeysets;
     }
 
     @Override
     public void upload(Collection<KeysetKey> data, JsonObject extraMeta) throws Exception {
+        if(!enableKeysets) {
+            LOGGER.error("Uploaded Attempted to Keysets Keys with keysets disabled");
```
Ian-Nara, line 34: I am confused about how the PrivateSiteDataSyncJob.java will work. We still execute the keysetSyncJob and keysetKeySyncJob and the writer provided by the factory handles "disabled" but the reader does not. So it will still read, compare, and upload but get stopped at at this line, right? Why even start the sync jobs if keysets are disabled?

src/main/java/com/uid2/admin/store/factory/KeysetKeyStoreFactory.java
```diff
@@ -22,20 +22,23 @@ public class KeysetKeyStoreFactory implements StoreFactory<Collection<KeysetKey>
...
 
     public KeysetKeyStoreFactory(
             ICloudStorage fileStreamProvider,
             CloudPath rootMetadataPath,
             VersionGenerator versionGenerator,
             Clock clock,
-            FileManager fileManager) {
+            FileManager fileManager,
+            boolean enableKeyset) {
         this.fileStreamProvider = fileStreamProvider;
```
jon8787, line 34: can we remove all these `enableKeyset` members and parameters and instead have a public static somewhere which returns `config.getBoolean("enable_keysets")` ?

src/test/java/com/uid2/admin/vertx/EncryptionKeyServiceKeysetsDisabledTest.java
```diff
@@ -0,0 +1,660 @@
...
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
+public class EncryptionKeyServiceKeysetsDisabledTest extends ServiceTestBase {
```
jon8787, line 27: - I guess all these copied tests are ok since it's temporary code which we'll remove in a couple months?
Otherwise we should use a technique like EUIDOperatorVerticleTest where the entire test suite is run again but with minor diffs for EUID.
- Are there any tests where we confirm the files are not created when they otherwise would be? Can we do this after the end of every test in this file, eg with some JUnit technique?

src/main/java/com/uid2/admin/Main.java
```diff
@@ -111,11 +112,13 @@ public void run() {
             CloudPath keysetMetadataPath = new CloudPath(config.getString(Const.Config.KeysetsMetadataPathProp));
             GlobalScope keysetGlobalScope = new GlobalScope(keysetMetadataPath);
             RotatingKeysetProvider keysetProvider = new RotatingKeysetProvider(cloudStorage, keysetGlobalScope);
-            KeysetStoreWriter keysetStoreWriter = new KeysetStoreWriter(keysetProvider, fileManager, jsonWriter, versionGenerator, clock, keysetGlobalScope);
+            KeysetStoreWriter keysetStoreWriter = new KeysetStoreWriter(keysetProvider, fileManager, jsonWriter, versionGenerator, clock, keysetGlobalScope, enableKeysets);
             try {
                 keysetProvider.loadContent();
```
jon8787, line 117: Why do we put `if(!enableKeysets)` in the catch block instead of around the whole try/catch?

src/main/java/com/uid2/admin/store/writer/KeysetKeyStoreWriter.java
cody-constine-ttd: Ok I added a if blocks to stop the job from getting executed.

src/test/java/com/uid2/admin/vertx/EncryptionKeyServiceKeysetsDisabledTest.java
cody-constine-ttd: I thought about making it parameterized, but I think for this being temporary and a test, its less effort and will be removed.

Added
```java
verifyNoInteractions(keysetStoreWriter);
verifyNoInteractions(keysetKeyStoreWriter);
```
To the end of every test in the disabled one.

src/main/java/com/uid2/admin/Main.java
cody-constine-ttd: Because I was over thinking it

src/main/java/com/uid2/admin/store/factory/KeysetKeyStoreFactory.java
cody-constine-ttd: I am not sure I  understand why this is better? Or how to implement a static that gets a config value.

src/test/java/com/uid2/admin/vertx/EncryptionKeyServiceKeysetsDisabledTest.java
jon8787: let's use `@AfterEach` instead if that works

src/main/java/com/uid2/admin/store/factory/KeysetKeyStoreFactory.java
jon8787: As discussed, if it's easy to do let's make a singleton so we can get config from anywhere (similar to uid2-core). Otherwise let's leave this for now.

jon8787 (approved):
Approved but left a couple comments for minor changes if they're easy to do

src/main/java/com/uid2/admin/job/jobsync/PrivateSiteDataSyncJob.java
```diff
@@ -166,14 +158,25 @@ public void execute() throws Exception {
...
-        KeysetKeySyncJob keysetKeySyncJob = new KeysetKeySyncJob(globalOperators, globalKeysetKeys, globalKeysets, globalMaxKeysetKeyId, keysetKeyWriter);
-
         siteSyncJob.execute();
         clientSyncJob.execute();
         encryptionKeySyncJob.execute();
         keyAclSyncJob.execute();
-        if(config.getBoolean("enable_keysets")) {
+        if(config.getBoolean(enableKeysetConfigProp)) {
+            Map<Integer, Keyset> globalKeysets = keysetStoreFactory.getGlobalReader().getSnapshot().getAllKeysets();
+            Collection<KeysetKey> globalKeysetKeys = keysetKeyStoreFactory.getGlobalReader().getSnapshot().getActiveKeysetKeys();
```
gmsdelmundo, line 167: Do we only want to sync active keysets?

PR #107, opened 2023-08-02
Ian UI2D 1573 1568 admin null allowlist and only one keyset

Ian-Nara:
Changes:

- Return an error if you try to make a keyset for a site that already has one (added test for this and commented out the test for allowing site to have multiple keysets as long as they have different names).
- Added a checkbox to the UI for setting the allowed_sites to null when "update" is clicked.
- You can also use "make new keyset" to create an empty, null, or a non-empty allowlist without having to make a second api call now.
- Some reorganization of the logic in SharingService.java#handleSetKeyset.
- KeysetKeys & SiteKeys are displayed sorted by id


webroot/adm/keysets.html
```diff
@@ -118,7 +118,12 @@ <h3>Output</h3>
 
         let payload = {keyset_id: keysetId}
 
-        if((add && add.length !== 0) || (remove && remove.length !== 0)) {
+        if ($('#nullify').is(':checked')){
+          allowedSites = null
+        } else if((add && add.length !== 0) || (remove && remove.length !== 0)) {
+          if(allowedSites !== undefined && allowedSites == null) {
+            $('#errorOutput').html('cannot set siteIds for null allowed_sites keyset. Please reset to [] first')
```
jon8787, line 125: - I find the "Reset Keyset" terminology quite cryptic. Can we change it to something like "Reset allowed_sites to empty"?
- Also the UX for resetting to empty is quite different from resetting to null. Do we want to make them more similar?

src/test/java/com/uid2/admin/vertx/SharingServiceTest.java
```diff
@@ -559,37 +558,38 @@ void KeysetSetBadWhitelistSiteId(Vertx vertx, VertxTestContext testContext) {
...
-        post(vertx, "api/sharing/keyset", body, ar -> {
-            assertTrue(ar.succeeded());
-            HttpResponse response = ar.result();
-            assertEquals(400, response.statusCode());
-            assertEquals("Keyset with same site_id and name already exists", response.bodyAsJsonObject().getString("message"));
-
-            testContext.completeNow();
-        });
-    }
+    // This test should be uncommented out when multiple keysets is enabled
```
jon8787, line 561: will @Disabled work here instead of commenting it out?

webroot/adm/keysets.html
jon8787: I've changed this to a ticket UID2-1598 so that we can close this PR. 
This PR looks good - please merge!

jon8787 (approved):

src/test/java/com/uid2/admin/vertx/SharingServiceTest.java
Ian-Nara: switched this to disabled

PR #113, opened 2023-08-12
UID2-882 site domain names

Ian-Nara:
[UID2-882](https://atlassian.thetradedesk.com/jira/browse/UID2-882)
add domain name functionality to SiteService

Note that currently, the first time sites.json is written to, all sites without domain_names field will get domain_names: []
have tested with local-config.json using postman & cloudberry

you will need to install this shared branch before testing: ian-UID2-882-site-domain-names


sunnywu:
looks ok after quick look but will wait for your admin front end changes and i'll give it all a manual test first

src/main/java/com/uid2/admin/vertx/service/SiteService.java
```diff
@@ -196,4 +208,75 @@ private void handleSiteEnable(RoutingContext rc) {
...
+
+            storeWriter.upload(sites, null);
+
+            rc.response().end(jsonWriter.writeValueAsString(existingSite));
+        } catch (Exception e) {
+            ResponseUtil.errorInternal(rc, "set site domain_names failed", e);
+        }
+    }
+
+    public static String getTopLevelDomainName(String origin) throws MalformedURLException {
```
sunnywu, line 261: maybe we should raise a new PR in UID2 shared lib with this function added so both operator and admin service can use it and we use the same single method to get domain name to be consistent?

src/main/java/com/uid2/admin/vertx/service/SiteService.java
```diff
@@ -196,4 +208,75 @@ private void handleSiteEnable(RoutingContext rc) {
...
+        String host;
+        try {
+            URL url = new URL(origin);
+            host = url.getHost();
+        } catch (Exception e) {
+            host = origin;
+        }
+        //InternetDomainName will normalise the domain name to lower case already
+        InternetDomainName name = InternetDomainName.from(host);
```
sunnywu, line 270: i'm surprised this works without changing pom.xml to explicitly add a google guava dependency?

sunnywu:
it seems there is some build issue above ?

src/main/java/com/uid2/admin/vertx/service/SiteService.java
Ian-Nara: hmm it works fine for me without it but I will add it in if that's just me.

src/main/java/com/uid2/admin/vertx/service/SiteService.java
Ian-Nara: the admin ui can expect/accpet domains that are passed as top-level domains that are not valid urls. like considering asfd.net valid instead of expecting it to be input as https://asdf.net. So the logic is a bit different I think.  If operator is fine to use this same behavior then may as well move it, otherwise I think the method small enough to just keep here.

Ian-Nara:
> it seems there is some build issue above ?

I added guava as a dependency. Also make sure you have checked out my shared branch  ian-UID2-882-site-domain-names. lmk if still problem.

src/main/java/com/uid2/admin/vertx/service/SiteService.java
Ian-Nara: ok, here is a ticket for this. Let me know if the description looks correct. [UID2-1659](https://atlassian.thetradedesk.com/jira/browse/UID2-1659)

sunnywu: looks good thanks! putting it on top of backlog for now. Please work on this when we merge admin/operator back to master

src/main/java/com/uid2/admin/vertx/service/SiteService.java
Ian-Nara: I accidentally put the google dependency in the ui branch. I have added it to this branch now and also merged master into my shared branch. No build issues for me

sunnywu (approved):

sunnywu:
it would be best to add a bit more logic in com.uid2.admin.vertx.service.SiteService#handleSiteAdd so people can specify the domain names directly when creating a new site without doing Set Domain Names as an explicit second/separate step. 

If you could do that and add 2 tests - 1 test site creation without domain name (should have a test already?) and another test specifying domain names then this PR should be good. Up to you if you want me to review adding this or otherwise you can merge everything to master starting with uid2-shared [PR ](https://github.com/IABTechLab/uid2-shared/pull/107/files)first

PR #126, opened 2023-08-27
add cstg key to trivy-secret.yaml

Ian-Nara:
[UID2-1703](https://atlassian.thetradedesk.com/jira/browse/UID2-1703)

### DIFF
--- a/trivy-secret.yaml
+++ b/trivy-secret.yaml
@@ -195,34 +195,6 @@ rules:
     regex: EUID-O-P-[0-9]*-(?P<secret>.{6}\..{38})
     secret-group-name: secret
 
-  ########################################
-  # EUID Client Side Keypair Private Key #
-  ########################################
-  - id: euid-client-side-keypair-private-key-test
-    category: euid
-    title: EUID - Client Side Keypair Private Key - Test
-    severity: CRITICAL
-    keywords:
-      - EUID-Y-T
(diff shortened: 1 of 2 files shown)


trivy-secret.yaml
```diff
@@ -83,6 +83,34 @@ rules:
...
+  ########################################
+  # UID2 Client Side Keypair Private Key #
+  ########################################
+  - id: uid2-client-side-keypair-private-key-test
+    category: uid2
+    title: UID2 - Client Side Keypair Private Key - Test
+    severity: CRITICAL
+    keywords:
+      - UID2-Y-T
+    regex: UID2-Y-T-[0-9]*-(?P<secret>.{92})
```
mcollins-ttd, line 95: Why the `-[0-9]*`? It's not in https://github.com/IABTechLab/uid2-optout/pull/42/files

Ian-Nara: Some keys in admin are prefixed like UID2-C-T-999-secret, so the other keys in trivy-secret.yaml are using this regex pattern. Im am not sure why.

mcollins-ttd: Ah ok, it looks like that number is the site ID. Keypairs don't have the site ID in their names.

mcollins-ttd (approved):

PR #130, opened 2023-08-30
Cbc UI d2 1177 keyset share types

cody-constine-ttd:


src/main/java/com/uid2/admin/Main.java
```diff
@@ -228,6 +228,14 @@ public void run() {
             RotatingKeysetProvider keysetProvider = new RotatingKeysetProvider(cloudStorage, keysetGlobalScope);
             KeysetStoreWriter keysetStoreWriter = new KeysetStoreWriter(keysetProvider, fileManager, jsonWriter, versionGenerator, clock, keysetGlobalScope, enableKeysets);
 
+            if(enableKeysets) {
```
jon8787, line 231: Can you add a large comment here about why it's important that this code had to be moved? What bad thing happens in the original place etc. I'd hope for a paragraph or so.

src/test/java/com/uid2/admin/job/jobsync/keyset/TestReplaceSharingtypesWithSitesJob.java
```diff
@@ -41,7 +41,7 @@ public class TestReplaceSharingtypesWithSitesJob {
     GlobalScope siteScope = new GlobalScope(globalSiteMetadataPath);
 
     List<Site> sites = List.of(
-            new Site(3, "site 3", true,  new HashSet<>()),
+            new Site(3, "site 3", true),
```
jon8787, line 44: can remove the "new HashSet<>" arguments below

src/main/java/com/uid2/admin/vertx/service/SiteService.java
```diff
@@ -112,7 +112,7 @@ private void handleSiteList(RoutingContext rc) {
                 jo.put("id", site.getId());
                 jo.put("name", site.getName());
                 jo.put("enabled", site.isEnabled());
-                jo.put("types", site.getClientTypes());
+                jo.put("Client Types", site.getClientTypes());
```
jon8787, line 115: is this capitalization intentional and identical to the other place? where is that other place btw -  should we be deduplicating?

jon8787 (changes requested):

webroot/adm/site.html
```diff
@@ -81,9 +81,13 @@ <h3>Output</h3>
         $('#doSetTypes').on('click', function () {
             var siteId = encodeURIComponent($('#siteId').val());
             var types = encodeURIComponent($('#types').val());
+            if(types.length < 1) {
+                $('#errorOutput').html('Can not set types with empty types');
```
jon8787, line 85: what if you accidentally changed an empty types to say DSP and you want to undo that?

src/main/java/com/uid2/admin/job/jobsync/keyset/ReplaceSharingTypesWithSitesJob.java
```diff
@@ -57,15 +57,15 @@ public ReplaceSharingTypesWithSitesJob(JsonObject config, WriteLock writeLock) {
 
         AdminKeysetStoreFactory adminKeysetStoreFactory = new AdminKeysetStoreFactory(
                 cloudStorage,
-                new CloudPath(config.getString(Const.Config.KeysetsMetadataPathProp)),
+                new CloudPath(config.getString("admin_keysets_metadata_path")),
```
gmsdelmundo, line 60: Should this be added to `Const.Config`?

src/main/java/com/uid2/admin/vertx/service/SiteService.java
```diff
@@ -207,6 +207,9 @@ private void handleSiteTypesSet(RoutingContext rc) {
             }
 
             Set<ClientType> types = getTypes(rc.queryParam("types").get(0));
+            if(types == null) {
+                ResponseUtil.error(rc, 400, "Invalid Types");
```
aulme, line 211: Should we return at this point?

src/main/java/com/uid2/admin/job/jobsync/keyset/ReplaceSharingTypesWithSitesJob.java
aulme: I would rather not do Const if we're not using this property elsewhere. Ideally this should be done with deserialization though

webroot/adm/site.html
aulme: Perhaps we can use a `confirm` dialog? I know it adds to the 90s vibe of the website but we're kinda already there.
I imagine most of the time people not putting types is an error but occasionally it's intentional as Jonathan mentioned.

src/main/java/com/uid2/admin/job/jobsync/keyset/ReplaceSharingTypesWithSitesJob.java
jon8787: Rather, should we remove everything from Const.Config that's only being used once?

cody-constine-ttd: Not in the PR Jonathan

src/test/java/com/uid2/admin/job/jobsync/keyset/TestReplaceSharingtypesWithSitesJob.java
cody-constine-ttd: No this was the thing I was explaining about how java is being weird about overloading the constructors.
Me and Ian both added new Set but his is a Set<String> and mine is a Set<ClientType> but for whatever reason Java doesn't look at generic type of the Set in the parameter, so we can't have two constructors with only 4 parameters.

webroot/adm/site.html
cody-constine-ttd: Added a confirm with a message warning about what will happen

jon8787 (approved):

PR #132, opened 2023-08-30
Ian UID- 1737 services and service_links UI

Ian-Nara:
[UID2-1737](https://atlassian.thetradedesk.com/jira/browse/UID2-1737)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
         <!-- check micrometer.version vertx-micrometer-metrics consumes before bumping up -->
         <micrometer.version>1.1.0</micrometer.version>
         <junit-jupiter.version>5.7.0</junit-jupiter.version>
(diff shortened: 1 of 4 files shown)


asloobq (commented):
I am curious where the request validation will occur example when creating a service
1] Site Id exists
2] Service Id is not duplicate
3] Service name is not empty and unique
3] Roles are valid

Ian-Nara:
> I am curious where the request validation will occur example when creating a service 1] Site Id exists 2] Service Id is not duplicate 3] Service name is not empty and unique 3] Roles are valid

These kind of validation are done in inside ServiceService.java and in ServiceLinkService.java. I have not enforced non-empty and unique name but I will add this.

thomasm-ttd (approved):

PR #133, opened 2023-08-31
Add sharer role and types to the site

jingyi-gao-ttd:


src/main/resources/localstack/s3/clients/clients.json
```diff
@@ -50,7 +50,8 @@
     "contact": "Publisher",
     "created": 1609459200,
     "roles": [
-      "GENERATOR"
+      "GENERATOR",
+      "SHARER"
```
jon8787, line 54: Do we need this? Publishers generally won't have SHARER to start with

lionell-pack-ttd: As I understand it, any participant type can have the SHARER permission - is there some reason this won't work? Afaik once these changes go live, a publisher would be able to include another publisher in their sharing list.

jon8787: Yes it will work, but it's a very uncommon scenario. What I'm getting at is that we should keep a GENERATOR-only example in this file, as that's a much more common case

src/main/resources/localstack/s3/sites/sites.json
```diff
@@ -2,27 +2,32 @@
...
-    "enabled": true
+    "enabled": true,
+    "clientTypes": ["DATA_PROVIDER"]
   },
   {
     "id": 126,
     "name": "Advertiser",
-    "enabled": true
+    "enabled": true,
+    "clientTypes": ["ADVERTISER"]
```
jon8787, line 30: I think we need some sites that have no clientTypes assigned yet, right?

jingyi-gao-ttd: Yea, good point! sites added

jon8787 (approved):

lionell-pack-ttd (approved):

PR #147, opened 2023-09-14
Added the ability to set and update service_id

thomasm-ttd:


src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -437,4 +447,17 @@ private void handleClientRename(RoutingContext rc) {
             rc.fail(500, e);
         }
     }
+
+    private int getServiceId(RoutingContext rc) {
+        Integer serviceId = 0;
```
gmsdelmundo, line 452: This can be `int` since we don't want null `serviceId` anyway

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -293,8 +297,14 @@ private void handleClientUpdate(RoutingContext rc) {
             final Site site = RequestUtil.getSite(rc, "site_id", this.siteProvider);
             if (site == null) return;
 
+            final int serviceId = this.getServiceId(rc);
+
             existingClient.withSiteId(site.getId());
 
+            if (serviceId != 0) {
```
gmsdelmundo, line 304: Will there be a case where we do want to convert a client back to 0 service ID? e.g. accidental update

thomasm-ttd: Good point. Updated

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
thomasm-ttd: Done

gmsdelmundo (approved):

PR #152, opened 2023-09-22
Operator plaintext key field removal

gmsdelmundo:


vishalegbert-ttd (approved):

webroot/adm/client-key.html
```diff
@@ -82,17 +82,6 @@ <h3>Output</h3>
             doApiCall('POST', url, '#standardOutput', '#errorOutput');
         });
 
-        $('#doDel').on('click', function () {
```
vishalegbert-ttd, line 85: Why?

src/main/java/com/uid2/admin/vertx/service/ClientJsonParser.java
```diff
@@ -0,0 +1,24 @@
...
+import com.uid2.shared.util.Mapper;
+import io.vertx.core.json.JsonObject;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Arrays;
+import java.util.Collection;
+
+// UID2-1503: Temporary for backwards compatibility
+public class ClientJsonParser implements Parser<Collection<JsonObject>> {
```
vishalegbert-ttd, line 16: Still needed?

webroot/adm/client-key.html
gmsdelmundo: We don't support client delete anymore, should we still keep it?

src/main/java/com/uid2/admin/vertx/service/ClientJsonParser.java
gmsdelmundo: Nope, deleted

PR #158, opened 2023-09-29
Add a new v2 router section.

lionell-pack-ttd:
Introduce attribute-based routing configuration.
Add a router to handle a site-specific request for client-side keypairs.


lionell-pack-ttd:
Branch name was wrong - I'll re-submit.

PR #159, opened 2023-09-29
Add a new v2 router section.

lionell-pack-ttd:
Introduce attribute-based routing configuration.
Add a router to handle a site-specific request for client-side keypairs.

### DIFF
--- a/src/main/java/com/uid2/admin/vertx/api/V2RouterModule.java
+++ b/src/main/java/com/uid2/admin/vertx/api/V2RouterModule.java
@@ -1,0 +1,29 @@
+package com.uid2.admin.vertx.api;
+
+import com.uid2.admin.secret.IKeypairManager;
+import com.uid2.admin.vertx.api.cstg.GetClientSideKeypairsBySite;
+import com.uid2.shared.middleware.AuthMiddleware;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class V2RouterModule {
+    private static final Logger LOGGER = LoggerFactory.getLogger(V2RouterModule.class);
+
+    private final IKeypairManager keypairManager;
+    private final AuthMiddleware authMiddleware;
+
+    public V2RouterModule(IKeypairManager keypairManager, AuthMiddleware authMiddleware) {
+        this.keypairManager = keypairManager;
+        this.authMiddleware = authMiddleware;
+    }
+
+    protected IRouteProvider[] getRouteProviders() {
+        return new IRouteProvider[] {
+                new GetClientSideKeypairsBySite(keypairManager)
+        };
+    }
+
+    public V2Router getRouter() {
+        return new V2Router(getRouteProviders(), authMiddleware);
+    }
+}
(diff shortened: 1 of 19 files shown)


src/main/java/com/uid2/admin/vertx/api/annotations/ApiMethod.java
```diff
@@ -0,0 +1,12 @@
+package com.uid2.admin.vertx.api.annotations;
+
+import io.vertx.core.http.HttpMethod;
+
+public enum ApiMethod {
+    GET(HttpMethod.GET);
```
jingyi-gao-ttd, line 6: I am not familiar with this, what does this line mean?

lionell-pack-ttd: That's passing HttpMethod.GET to the enum constructor for GET.

If the enum had no constructor, it would just be:
`public enum ApiMethod {
  GET;
}

But because there's a constructor which requires a `HttpMethod` argument, instead of just `GET` you need to pass it the matching `HttpMethod`.

It's because I wanted to use the HTTP method in an attribute, but unfortunately Vert.x's methods aren't compile-time constants 😥 so this enum gives a way to match.

Once someone needs `POST` (or another verb) it should be obvious enough how to add it.

jingyi-gao-ttd: Oh, I see, I didn't realise this ApiMethod is an enum! the `HttpMethod.GET` is the vertxMethod that passes into the constructor now it makes sense to me. Thanks for the explanation.

src/main/java/com/uid2/admin/vertx/api/V2RouterModule.java
```diff
@@ -0,0 +1,29 @@
...
+    private final AuthMiddleware authMiddleware;
+
+    public V2RouterModule(IKeypairManager keypairManager, AuthMiddleware authMiddleware) {
+        this.keypairManager = keypairManager;
+        this.authMiddleware = authMiddleware;
+    }
+
+    protected IRouteProvider[] getRouteProviders() {
+        return new IRouteProvider[] {
+                new GetClientSideKeypairsBySite(keypairManager)
```
jingyi-gao-ttd, line 22: I understand this is a workaround after taking out Guice for now. Now if we want to add a new v2 endpoint, we will need to update the constructor function, pass the new service into the constructor, and add the new route provider here. I wonder if this intent to keep those workarounds in one file? 

I wonder if it will be better for now to add a new interface (IV2Service ? ) to services with an abstract function like `getV2RouteProviders`? . And then the constructor took an array of IV2Service and for each one called `getV2RouteProviders`? Thus, before the DI merged in, we could change the service to support the V2 Routes?

lionell-pack-ttd: I'm trying to limit the reasons `Main` needs to change - right now it changes for basically any change in the relationships between any two classes in the whole system. At least this way, it only needs to change if something in the V2 routes needs a new service. I'm hoping we can introduce DI fairly soon, but this limits the problem.

If you need to write a new route which only uses services that are already used in the V2 module, no changes are needed in `Main`, which I think is a small step forward.

PR #160, opened 2023-10-01
backfill site created timestamp from oldest client key

Ian-Nara:
tested locally and with unit test.

My plan is to deploy from this branch and trigger the update for each of the sites.json files, then downgrade admin back to previous version and close this PR.

### DIFF
--- a/src/main/java/com/uid2/admin/vertx/service/SiteService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/SiteService.java
@@ -72,15 +72,41 @@ public class SiteService implements IService {
             }
         }, Role.CLIENTKEY_ISSUER));
         router.post("/api/site/set-types").blockingHandler(auth.handle((ctx) -> {
-                    synchronized (writeLock) {
-                        this.handleSiteTypesSet(ctx);
-                    }
-                    }, Role.CLIENTKEY_ISSUER));
+            synchronized (writeLock) {
+                this.handleSiteTypesSet(ctx);
+            }
+        }, Role.CLIENTKEY_ISSUER));
         router.post("/api/site/domain_names").blockingHandler(auth.handle((ctx) -> {
             synchronized (writeLock) {
                 this.handleSiteDomains(ctx);
             }
         }, Role.CLIENTKEY_ISSUER));
+        router.post("/api/site/temporary/backfill_created").blockingHandler(auth.handle((ctx) -> {
+            synchronized (writeLock) {
+                this.handleSiteBackfillCreated(ctx);
+            }
+        }, Role.CLIENTKEY_ISSUER));
+    }
+
+    private void handleSiteBackfillCreated(RoutingContext rc) {
+        try {
+            final Collection<Site> sites = new ArrayList<>(this.siteProvider.getAllSites());
+            final Collection<Site> updatedSites = new ArrayList<>();
+            for (Site s : sites) {
+                ClientKey key = this.clientKeyProvider.getOldestClientKey(s.getId());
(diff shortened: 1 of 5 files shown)


src/main/java/com/uid2/admin/vertx/service/SiteService.java
```diff
@@ -72,15 +72,41 @@ public void setupRoutes(Router router) {
...
+            synchronized (writeLock) {
+                this.handleSiteTypesSet(ctx);
+            }
+        }, Role.CLIENTKEY_ISSUER));
         router.post("/api/site/domain_names").blockingHandler(auth.handle((ctx) -> {
             synchronized (writeLock) {
                 this.handleSiteDomains(ctx);
             }
         }, Role.CLIENTKEY_ISSUER));
+        router.post("/api/site/temporary/backfill_created").blockingHandler(auth.handle((ctx) -> {
```
cody-constine-ttd, line 84: Why do we need an endpoint for this? Can't admin just do it on start up?

Ian-Nara: Changed this so it is done when SiteService setupRoutes() is called and removed the unit test. Checked it still works running locally

src/main/java/com/uid2/admin/vertx/service/SiteService.java
```diff
@@ -72,15 +72,36 @@ public void setupRoutes(Router router) {
...
             synchronized (writeLock) {
                 this.handleSiteDomains(ctx);
             }
         }, Role.CLIENTKEY_ISSUER));
+        handleSiteBackfillCreated();
+    }
+
+    private void handleSiteBackfillCreated() {
+        try {
+            final Collection<Site> sites = new ArrayList<>(this.siteProvider.getAllSites());
```
asloobq, line 89: Just curious why this needs to be mapped to an ArrayList ? `this.siteProvider.getAllSites()` returns a `Collection<Site>` which is iterable in a for loop.

src/main/java/com/uid2/admin/vertx/service/SiteService.java
```diff
@@ -72,15 +72,36 @@ public void setupRoutes(Router router) {
...
+            final Collection<Site> updatedSites = new ArrayList<>();
+            for (Site s : sites) {
+                ClientKey key = this.clientKeyProvider.getOldestClientKey(s.getId());
+                if (s.getCreated() == 0 && key != null) {
+                    long created = key.getCreated();
+                    Site us = new Site(s.getId(), s.getName(), s.isEnabled(), s.getClientTypes(), s.getDomainNames(), created);
+                    updatedSites.add(us);
+                } else {
```
asloobq, line 97: Do you think its worth it to log sites where the `created` timestamp is 0 as well as there are no valid client keys ? Or you have from production sites.json that this scenario doesn't exist ?

asloobq (commented):
Some nitpicking comments, may not be important if code is going to run only once.

src/main/java/com/uid2/admin/vertx/service/SiteService.java
Ian-Nara: fixed

src/main/java/com/uid2/admin/vertx/service/SiteService.java
Ian-Nara: It may be possible for sites to exist without a client key like if one created one in testing, but since we can't delete sites I don't think there is a point in listing them.

asloobq (approved):

cody-constine-ttd (approved):

PR #161, opened 2023-10-03
Use `uid2-mock-data` as resource files in admin

cYKatherine:


lionell-pack-ttd:
This removes a lot of test data which we're using to work on integrations with the portal, and takes away our current primary way of running the portal. I expect it would take a *lot* of discussion and work to merge this! Where will this data come from now? What work has been done to match the other source of test data up with the values that you're removing? How do we support some of the developer workflows we're currently using this for (e.g. testing the UID2 Portal)?

src/main/resources/localstack/s3/sites/sites.json
```diff
@@ -1,57 +0,0 @@
...
-    "enabled": true,
-    "clientTypes": ["ADVERTISER"]
-  },
-  {
-    "id": 127,
-    "name": "OptOut",
-    "enabled": true
-  },
-  {
-    "id": 128,
```
lionell-pack-ttd, line 38: Some of these are missing from the new mock data repo (e.g. 128-130) - I'm not sure how much other devs are relying on the test data that has been added here, but I know we've done some work to make sure the test data here matches other systems.

I haven't checked other JSON files, but we probably need to make sure the new test data is a superset of the various copies we're combining, otherwise it'll cause people a bunch of work to debug why their test data suddenly stopped working.

lionell-pack-ttd:
Should the mock data repo be in the IABTechLab org? It seems unusual that it's not where all the other repos are.

PR #163, opened 2023-10-04
Edit service and service links

thomasm-ttd:
Edit and delete Service and Service_links


src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
```diff
@@ -118,6 +128,102 @@ private void handleServiceLinkAdd(RoutingContext rc) {
...
+            }
+
+            if (siteProvider.getSite(siteId) == null) {
+                ResponseUtil.error(rc, 404, "site_id " + siteId + " not valid");
+                return;
+            }
+
+            ServiceLink serviceLink = this.serviceLinkProvider.getAllServiceLinks()
+                    .stream().filter(s -> s.getServiceId() == serviceId && s.getSiteId() == siteId)
+                    .findFirst()
```
Ian-Nara, line 162: If we are allowing multiple service_links to a service, will this be enough to identify the service_link

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
```diff
@@ -118,6 +128,102 @@ private void handleServiceLinkAdd(RoutingContext rc) {
...
+            Integer serviceId = body.getInteger("service_id");
+
+            if (linkId == null || serviceId == null) {
+                ResponseUtil.error(rc, 400, "required parameters: link_id, service_id");
+                return;
+            }
+
+            ServiceLink serviceLink = this.serviceLinkProvider.getAllServiceLinks()
+                    .stream().filter(s -> s.getServiceId() == serviceId && s.getLinkId().equals(linkId))
+                    .findFirst()
```
Ian-Nara, line 208: If we are allowing multiple service_links to a service, will this be enough to identify the service_link

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
thomasm-ttd: Good catch - update to include the name, so the only thing you can change is the LinkId. As we can delete a ServiceLink, it is easy to recreate it if we need to.

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
thomasm-ttd: The linkId must be unique within a service. When authorizing a caller, we only use the LinkId and the service, so we don't want to allow duplicates. On create, we check it is unique on those 2 properties.

src/main/java/com/uid2/admin/vertx/service/ServiceService.java
```diff
@@ -169,26 +176,72 @@ private void handleServiceRoles(RoutingContext rc) {
...
+                try {
+                    roles = rolesSpec.stream().map(s -> Role.valueOf((String) s)).collect(Collectors.toSet());
+                } catch (IllegalArgumentException e) {
+                    ResponseUtil.error(rc, 400, "invalid parameter: roles");
+                    return;
+                }
+                service.setRoles(roles);
+            }
+
+            if (siteId != null && siteId != 0) {
```
asloobq, line 190: I think we need to re-validate that the <service_name, site_id> combination remains unique

src/main/java/com/uid2/admin/vertx/service/ServiceService.java
```diff
@@ -169,26 +176,72 @@ private void handleServiceRoles(RoutingContext rc) {
...
-                return;
+            if (rolesSpec != null) {
+                final Set<Role> roles;
+                try {
+                    roles = rolesSpec.stream().map(s -> Role.valueOf((String) s)).collect(Collectors.toSet());
+                } catch (IllegalArgumentException e) {
+                    ResponseUtil.error(rc, 400, "invalid parameter: roles");
+                    return;
+                }
+                service.setRoles(roles);
```
asloobq, line 187: is `roles` guaranteed to be non-empty here?

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
```diff
@@ -118,6 +128,103 @@ private void handleServiceLinkAdd(RoutingContext rc) {
...
+            if (serviceLink == null) {
+                ResponseUtil.error(rc, 404, "failed to find a service_link for serviceId: " + serviceId + " and link_id: " + linkId);
+                return;
+            }
+
+            final List<ServiceLink> serviceLinks = this.serviceLinkProvider.getAllServiceLinks()
+                    .stream().sorted(Comparator.comparingInt(ServiceLink::getServiceId))
+                    .collect(Collectors.toList());
+
+            serviceLinks.remove(serviceLink);
```
asloobq, line 220: do we want to keep track of which services  and links were created and deleted in the past ? If that is useful then we should consider soft deletion / disabling them

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
thomasm-ttd: Having a bunch of disabled keys that we know are wrong doesn't seem like a good idea. Having the ability to edit, as well as delete, gives us the option to maintain the keys. We currently don't have the concept of disabled service_links either.

src/main/java/com/uid2/admin/vertx/service/ServiceService.java
Ian-Nara: We allow it to be non empty in other places like the client key. But if we are able to delete these then maybe it is a good idea to not allow roles to be non-empty, and instead delete the service if it can't do anything.

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
Ian-Nara: Ok, makes sense

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
```diff
@@ -54,6 +54,16 @@ public void setupRoutes(Router router) {
                 this.handleServiceLinkAdd(ctx);
             }
         }, Role.ADMINISTRATOR));
+        router.post("/api/service_link/update").blockingHandler(auth.handle((ctx) -> {
```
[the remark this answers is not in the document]
thomasm-ttd: I was following the conventions in the rest of the code base

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
```diff
@@ -118,6 +128,103 @@ private void handleServiceLinkAdd(RoutingContext rc) {
...
+                ResponseUtil.error(rc, 404, "service_id " + serviceId + " not valid");
+                return;
+            }
+
+            if (siteProvider.getSite(siteId) == null) {
+                ResponseUtil.error(rc, 404, "site_id " + siteId + " not valid");
+                return;
+            }
+
+            final List<ServiceLink> serviceLinks = this.serviceLinkProvider.getAllServiceLinks()
```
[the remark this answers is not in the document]
thomasm-ttd: Sorting them before write should make the retrieval slightly quicker

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
```diff
@@ -118,6 +128,103 @@ private void handleServiceLinkAdd(RoutingContext rc) {
...
+                ResponseUtil.error(rc, 404, "failed to find a service_link for serviceId: " + serviceId + ", site_id: " + siteId + " and name: " + name);
+                return;
+            }
+
+            if (serviceLinks.stream().anyMatch(sl -> sl.getServiceId() == serviceId && sl.getLinkId().equals(linkId))) {
+                ResponseUtil.error(rc, 400, "service link id already exists for service_id: " + serviceId);
+                return;
+            }
+
+            if (linkId != null && !linkId.isEmpty()) {
```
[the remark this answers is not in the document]
thomasm-ttd: Added it to the validation of the parameters

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
```diff
@@ -118,6 +128,103 @@ private void handleServiceLinkAdd(RoutingContext rc) {
...
+        } catch (Exception e) {
+            ResponseUtil.errorInternal(rc, "Internal Server Error", e);
+        }
+    }
+
+    private void handleServiceLinkDelete(RoutingContext rc) {
+        try {
+            serviceProvider.loadContent();
+            JsonObject body = rc.body().asJsonObject();
+            if (body == null) {
```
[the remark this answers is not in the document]
thomasm-ttd: 2 of the 4 parameters are shared, and the error message returned is different. I don't think factoring this into shared code will make it easier to understand or maintain.

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
thomasm-ttd: I'm not sure there is currently a need to keep track of disabled keys. A number of other entities also support deleting the key

src/main/java/com/uid2/admin/vertx/service/ServiceService.java
thomasm-ttd: Updated the validation of the roles and added some tests. We don't actually use roles on the service at the moment for actually validating what the service can do, so this is a future enhancement

src/main/java/com/uid2/admin/vertx/service/ServiceService.java
thomasm-ttd: Added the validation and a unit test

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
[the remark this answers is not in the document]
asloobq: I would like to see us move to REST APIs in the future. Probably need a larger discussion for this

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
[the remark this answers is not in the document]
asloobq: Okay. I didn't about this performance feature of streams

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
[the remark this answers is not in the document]
asloobq: Can you please point me to the code where this validation is done ?

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
asloobq: Ok, I feel a little uneasy about not leaving a trace of the deleted key and who deleted it. Could we at least write a log message to record this info ?

src/main/java/com/uid2/admin/vertx/service/ServiceService.java
asloobq: Oh I didn't know that. We should create a ticket to track this.

src/main/java/com/uid2/admin/vertx/service/ServiceService.java
asloobq: sorry, I couldn't find the validation. Can you please add a link to it ?

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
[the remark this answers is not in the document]
thomasm-ttd: I validate that the link_id is not null and not empty, and have updated the search to include the link_id in the search. If it is not found, it will return a 404

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
thomasm-ttd: I don't disagree, but there is no auditing anywhere in Admin, so I think this is a bigger conversation. We also want to move away from using Github as the authentication mechanism.

src/main/java/com/uid2/admin/vertx/service/ServiceService.java
thomasm-ttd: Ticket created: UID2-2084

asloobq (approved):
LGTM

PR #166, opened 2023-10-09
update site created timestamp

Ian-Nara:
tested locally and with unit test


asloobq (dismissed):

src/main/java/com/uid2/admin/vertx/service/SiteService.java
```diff
@@ -296,6 +301,40 @@ private void handleSiteDomains(RoutingContext rc) {
         }
     }
 
+    private void handleSiteCreated(RoutingContext rc) {
```
lionell-pack-ttd, line 304: I started out thinking this was handling a new site created action - changing the name to something like "handleUpdateSiteCreated" ior "handleUpdateCreatedTimestamp" would be clearer.

PR #170, opened 2023-10-16
Confirmations and basic validation when editing keys in UI

alex-yau-ttd:
- Used roughly existing pattern for confirmations and validation when submitting an action, mainly around disabling or enabling keys.
- Extracted validation where it made sense - though more effort needs to be done here across all the pages
- Added README section to help devs work on the UI

## Manual tests

**Disabling and enabling a client key**

https://github.com/IABTechLab/uid2-admin/assets/137864392/f8268cc3-11cf-450e-8460-7cdc98dcbbcf

**Basic validation**

https://github.com/IABTechLab/uid2-admin/assets/137864392/86325fa8-7f65-4a58-8000-382ec809c9b0

### DIFF
--- a/webroot/adm/administrators.html
+++ b/webroot/adm/administrators.html
@@ -25,7 +25,7 @@
 
 <ul>
     <li class="ro-adm" style="display: none"><a href="#" id="doMeta">Get Metadata</a></li>
-    <li class="ro-adm" style="display: none"><a href="#" id="doList">List Admin User</a></li>
+    <li class="ro-adm" style="display: none"><a href="#" id="doList">List Admin Users</a></li>
     <li class="ro-adm" style="display: none"><a href="#" id="doReveal">Reveal Admin User</a></li>
     <li class="ro-adm" style="display: none"><a href="#" id="doAdd">Add Admin User</a></li>
     <li class="ro-adm" style="display: none"><a href="#" id="doDisable">Disable Admin User</a></li>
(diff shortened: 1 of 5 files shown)


webroot/adm/client-side-keypairs.html
```diff
@@ -93,20 +93,35 @@ <h3>Output</h3>
...
             }
 
             doApiCall('POST', '/api/client_side_keypairs/update', '#standardOutput', '#errorOutput', JSON.stringify(payload));
         });
 
     });
+
+    function getUpdateKeypairConfirmationMessage(disabled, subscriptionId) {
+        let action = disabled ? 'disable' : 'enable';
+        let confirmationMessage = `You may ignore this warning if you are not changing the disabled status of this keypair.\n\nAre you sure you want to ${action} ${subscriptionId}?`;
```
alex-yau-ttd, line 116: The pattern for client-side-keypairs is a little different, since it just has a generic update function which is given a disabled flag, rather than explicit `enable` and `disable` buttons. Ideally this would line would not be necessary, but I'm not sure how much effort we should put into making this better right now.

webroot/adm/administrators.html
```diff
@@ -110,18 +110,40 @@ <h3>Output</h3>
...
-            var url = '/api/admin/disable?name=' + email;
+            var email = $('#email').val();
+            var encodedEmail = encodeURIComponent(email);
+            var url = '/api/admin/disable?name=' + encodedEmail;
+
+            if (!validateEmail(encodedEmail)) return;
+
+            const confirmationMessage = `Disabling this admin key will prevent the associated account from being to interact with the admin service.\n\nAre you sure you want to disable ${email}?`;
+
+            if (!confirm(confirmationMessage)) return;
```
cYKatherine, line 121: Just curious where is this `confirm()` function defined?

alex-yau-ttd: good question - it comes for free with Javascript Window. See:
-  https://www.w3schools.com/jsref/obj_window.asp
- https://www.geeksforgeeks.org/javascript-window-confirm-method/

alex-yau-ttd: I wasn't familiar with it before this ticket, but we are already using it elsewhere in the project

cYKatherine (approved):

PR #173, opened 2023-11-03
Allow updating new site fields in UI

alex-yau-ttd:
- New api endpoint `/api/site/update` which currently handles `description` and `visible`
- Allow `description` to be passed to add site method
- Add buttons to the UI to set description and visible 
- The first commit is purely formatting changes
- Relevant changes in shared: https://github.com/IABTechLab/uid2-shared/pull/157

## Local Testing

**Listing sites**

https://github.com/IABTechLab/uid2-admin/assets/137864392/b1e90df9-2c10-448f-9e71-b18cad30b907

**Adding a new site with a description**

https://github.com/IABTechLab/uid2-admin/assets/137864392/091db096-4251-452c-aa4f-8ae05b5ab986

**Adding a site with no description**

https://github.com/IABTechLab/uid2-admin/assets/137864392/7814cc9c-c99f-4147-9b94-4cabb04a4f74

**Updating a site**

https://github.com/IABTechLab/uid2-admin/assets/137864392/90d5f466-187d-429e-a3b3-9261d9e549c3

**Updating a site that has a null `visible`**

https://github.com/IABTechLab/uid2-admin/assets/137864392/96822ede-9e47-4ad0-a155-452ec7cf0503

### DIFF
--- a/src/main/java/com/uid2/admin/vertx/service/SiteService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/SiteService.java
@@ -81,6 +81,11 @@ public class SiteService implements IService {
                 this.handleSiteDomains(ctx);
             }
         }, Role.CLIENTKEY_ISSUER));
+        router.post("/api/site/update").blockingHandler(auth.handle((ctx) -> {
+            synchronized (writeLock) {
+                this.handleSiteUpdate(ctx);
+            }
+        }, Role.CLIENTKEY_ISSUER));
     }
 
     private void handleRewriteMetadata(RoutingContext rc) {
@@ -111,9 +116,11 @@ public class SiteService implements IService {
 
                 jo.put("id", site.getId());
                 jo.put("name", site.getName());
+                jo.put("description", site.getDescription());
                 jo.put("enabled", site.isEnabled());
                 jo.put("clientTypes", site.getClientTypes());
                 jo.put("domain_names", domainNamesJa);
+                jo.put("visible", site.isVisible());
                 jo.put("created", site.getCreated());
 
                 List<ClientKey> clients = clientKeys.getOrDefault(site.getId(), emptySiteKeys);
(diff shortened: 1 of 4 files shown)


src/main/java/com/uid2/admin/vertx/service/SiteService.java
```diff
@@ -296,11 +304,48 @@ private void handleSiteDomains(RoutingContext rc) {
...
+            if (description != null) {
+                existingSite.setDescription(description);
+            }
+            if (visibleParam != null) {
+                if ("true".equalsIgnoreCase(visibleParam)) {
+                    existingSite.setVisible(true);
+                } else if ("false".equalsIgnoreCase(visibleParam)) {
+                    existingSite.setVisible(false);
+                } else {
+                    ResponseUtil.error(rc, 400, "Invalid parameter for visible: " + visibleParam);
```
cYKatherine, line 328: I saw in your video when you `/list/sites`, some sites have `visible` set to null, I guess it's because we don't have that value in the json files in integ/prod environment. Should we update the json files for them? Otherwise would it throw a 400?

alex-yau-ttd: > I saw in your video when you /list/sites, some sites have visible set to null, I guess it's because we don't have that value in the json files in integ/prod environment.
yeah - to be clear, this was tested on local, so it's because the sites.json for localstack does not have a `visible` value. Indeed, all values in integ/prod environments will have a null `visible` field.

> Should we update the json files for them? Otherwise would it throw a 400?
I don't think this is strictly necessary today - the `visible` parameter is not technically used anywhere yet. Updating a site using this method will not throw a 400 if `visible` is null, since we check it above.

That all being said, it probably makes sense to update this value for existing sites integ/prod so it is accurate, but it should not break anything if we don't. 

What are your thoughts?

cYKatherine: Cool, if it doesn't break anything in integ and production with the null visible value I'm happy to see it as a separate task :D Can you create a ticket for that please?

cYKatherine (approved):

PR #177, opened 2023-11-21
Prevent NPE when updating allowed_sites from null

alex-yau-ttd:
`!existingAllowedSites.contains(s)` was causing an NPE when `existingAllowedSites` was `null`. This PR fixes that issue and adds a test for this situation.

**Manual tests in the Portal**
No longer receives a 500 from admin and crashes when `allowed_sites` is `null` - it now works as you would expect.

https://github.com/IABTechLab/uid2-admin/assets/137864392/4d0aed60-1d33-4b5c-8998-bc0155dc7a35

**Manual tests in Admin UI**
Preserves functionality.

https://github.com/IABTechLab/uid2-admin/assets/137864392/8d25e09e-5223-440e-a2c0-da710041640f

### DIFF
--- a/src/main/java/com/uid2/admin/vertx/service/SharingService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/SharingService.java
@@ -345,9 +345,11 @@ public class SharingService implements IService {
         final Set<Integer> newlist;
 
         if (allowedSites != null){
-            final Set<Integer> existingAllowedSites = existingSites;
+            final Set<Integer> existingAllowedSites = (existingSites != null) ? existingSites : new HashSet<>();
             OptionalInt firstInvalidSite = allowedSites.stream()
-                    .mapToInt(s -> (Integer) s).filter(s -> !existingAllowedSites.contains(s) && !isSiteIdEditable(s)).findFirst();
+                    .mapToInt(s -> (Integer) s)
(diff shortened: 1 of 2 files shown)


src/test/java/com/uid2/admin/vertx/SharingServiceTest.java
```diff
@@ -574,6 +610,7 @@ void KeysetSetNew(Vertx vertx, VertxTestContext testContext) {
         };
     }*/
 
+    @Test
```
alex-yau-ttd, line 613: @cody-constine-ttd I saw this test was missing the `@Test` annotation, so I've added it here

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -345,9 +345,11 @@ private AdminKeyset setAdminKeyset(RoutingContext rc, JsonArray allowedSites, Js
         final Set<Integer> newlist;
 
         if (allowedSites != null){
-            final Set<Integer> existingAllowedSites = existingSites;
+            final Set<Integer> existingAllowedSites = (existingSites != null) ? existingSites : new HashSet<>();
```
alex-yau-ttd, line 348: This is the actual fix

lionell-pack-ttd (approved):
Looks good to me.

cody-constine-ttd (approved):
LGTM

PR #185, opened 2023-12-06
UID2-988 - Moved optout mock data to uid2-optout

gmsdelmundo:


src/main/resources/localstack/init-aws.sh
```diff
@@ -1,10 +1,7 @@
 #!/usr/bin/env bash
 
 aws s3 --endpoint-url http://localhost:5001 mb s3://test-core-bucket
-aws s3 --endpoint-url http://localhost:5001 cp /s3/core/ s3://test-core-bucket/ --recursive
+aws s3 --endpoint-url http://localhost:5001 cp /s3/core/core/ s3://test-core-bucket/ --recursive
```
vishalegbert-ttd, line 4: Is this right?

src/main/java/com/uid2/admin/Main.java
```diff
@@ -311,12 +309,17 @@ public void run() {
...
             //UID2-575 set up a job dispatcher that will write private site data periodically if there is any changes
             //check job for every minute
-            PrivateSiteDataSyncJob job = new PrivateSiteDataSyncJob(config, writeLock);
-            jobDispatcher.enqueue(job);
-            jobDispatcher.executeNextJob();
-
+            PrivateSiteDataSyncJob privateSiteDataSyncJob = new PrivateSiteDataSyncJob(config, writeLock);
+            jobDispatcher.enqueue(privateSiteDataSyncJob);
+            CompletableFuture<Boolean> privateSiteDataSyncJobFuture = jobDispatcher.executeNextJob();
+            privateSiteDataSyncJobFuture.get();
```
vishalegbert-ttd, line 322: Previously it did not wait but now it will?

gmsdelmundo: Yep, this was basically the reason why we needed to wait 1min after starting admin/core before starting operators locally

If we call `executeNextJob()` back-to-back immediately, the second call won't work because the first job is still running. So you would have to wait 1min for the next job to be executed

src/main/resources/localstack/init-aws.sh
gmsdelmundo: It was needed to get both uid2-admin and uid2-optout's S3 initialisation scripts coexist, but I've realised I didn't have to make this change, so I'll be reverting this

vishalegbert-ttd (approved):

PR #189, opened 2023-12-12
sch-UID2-2470 add roles to service links

sophia-chen-ttd:
Added roles to service links in the ServiceLinkService and service-link.html.
Requirements for roles:
- Roles cannot be empty
- Roles must be valid and a subset of the roles in allowed in Service
- When removing a role in Service, it must not currently be in use by any of its ServiceLinks

Added unit tests and tested locally for:
- Adding service link with role not allowed by service should return an error
- Adding service link with illegal role should return an error
- Adding service link with no role when service allows for multiple roles will return an error
- Adding service link with no role when service allows for single role will automatically populate role
- Adding service link with roles allowed by service succeeds

- Updating service with name and no role succeeds, only name is changed
- Updating service with valid role and no name succeeds, only role is changed
- Updating service with role not allowed by service returns an error
- Updating service with both name and valid roles succeeds
- Updating service with no name and no role does not make any changes and does not return an error
- Removing role(s) from service that still exists in service links will return an error
- Removing role(s) from service that do not exist in service links will succeed


atarassov-ttd:
What should happen to existing service links (and their roles) if service roles are updated later?

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
```diff
@@ -107,11 +109,35 @@ private void handleServiceLinkAdd(RoutingContext rc) {
                 return;
             }
 
+            Set<Role> serviceRoles = serviceProvider.getService(serviceId).getRoles();
+            final Set<Role> roles;
+            if (rolesJson == null || rolesJson.isEmpty()) {
+                // if no roles provided and service only allows for one role, populate automatically
+                if (serviceRoles.size() == 1) {
```
atarassov-ttd, line 116: Why is the case of 1 service role treated specially? What's wrong with always defaulting to service roles?

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
```diff
@@ -92,6 +93,7 @@ private void handleServiceLinkAdd(RoutingContext rc) {
             Integer serviceId = body.getInteger("service_id");
             Integer siteId = body.getInteger("site_id");
             String name = body.getString("name");
+            JsonArray rolesJson = body.getJsonArray("roles");
             if (linkId == null || serviceId == null || siteId == null || name == null) {
```
atarassov-ttd, line 97: It looks like empty names are allowed here. Is that intentional?

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
```diff
@@ -172,7 +199,26 @@ private void handleServiceLinkUpdate(RoutingContext rc) {
...
             }
 
-            serviceLink.setName(name);
+            if (name != null && !name.isEmpty()) {
+                serviceLink.setName(name);
+            }
+
+            if (rolesJson != null && !rolesJson.isEmpty()) {
+                final Set<Role> roles;
+                try {
```
atarassov-ttd, line 208: This blocks looks very similar to the one in `handleServiceLinkAdd`: do you think it is possible/feasible to reuse the code?

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
sophia-chen-ttd: If a service allows for multiple roles, we want participants to specify which role they have for the service link as they may only have access to a subset of it. For example, if the snowflake service allows for MAPPER and SHARER roles, the user may only have access to the role MAPPER. So, we want them to have to specify which role they have rather than allowing them to default to all service roles.

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
sophia-chen-ttd: No, good catch! I'll add that in.

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
```diff
@@ -107,11 +109,33 @@ private void handleServiceLinkAdd(RoutingContext rc) {
...
             }
 
+            if (name.isEmpty()) {
+                ResponseUtil.error(rc, 400, "name cannot be empty");
+                return;
+            }
+
+            Set<Role> serviceRoles = serviceProvider.getService(serviceId).getRoles();
+            final Set<Role> roles;
+            if (rolesJson == null || rolesJson.isEmpty()) {
```
vishalegbert-ttd, line 119: We can reduce nesting in this block

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
```diff
@@ -107,11 +109,33 @@ private void handleServiceLinkAdd(RoutingContext rc) {
...
+                // if no roles provided and service only allows for one role, populate automatically
+                if (serviceRoles.size() == 1) {
+                    roles = Set.copyOf(serviceRoles);
+                } else {
+                    ResponseUtil.error(rc, 400, "required parameter: roles");
+                    return;
+                }
+            } else {
+                roles = getValidRoles(rolesJson, serviceRoles, rc);
+                if (roles == null) {
```
vishalegbert-ttd, line 129: This is weird, I think we can do better here

src/main/java/com/uid2/admin/vertx/service/ServiceService.java
```diff
@@ -203,7 +204,17 @@ private void handleUpdate(RoutingContext rc) {
                     ResponseUtil.error(rc, 400, "invalid parameter: roles");
                     return;
                 }
-                service.setRoles(roles);
+                // check if role is removed, that it is no longer in use by service links
+                Set<Role> removedRoles = service.getRoles().stream().filter(r -> !roles.contains(r)).collect(Collectors.toSet());
+                boolean removedRolesNotInUse = serviceLinkProvider.getAllServiceLinks().stream().filter(sl -> sl.getServiceId() == serviceId)
```
vishalegbert-ttd, line 209: This is a bit hard to read

src/test/java/com/uid2/admin/vertx/ServiceLinkServiceTest.java
```diff
@@ -52,7 +51,7 @@ private void checkServiceLinkJson(ServiceLink expectedServiceLink, JsonObject ac
     }
 
     @Test
-    void listLinksNoLinks(Vertx vertx, VertxTestContext testContext) {
+    void listLinks_NoLinks_ReturnsEmptyArray(Vertx vertx, VertxTestContext testContext) {
```
vishalegbert-ttd, line 54: Let's stay on camelCase 

Should be listLinks_noLinks_returnsEmptyArray

src/main/java/com/uid2/admin/Main.java
```diff
@@ -231,7 +232,7 @@ public void run() {
             JobDispatcher jobDispatcher = new JobDispatcher("job-dispatcher", 1000 * 60, 3, clock);
             jobDispatcher.start();
 
-            val clientSideKeypairService = new ClientSideKeypairService(config, auth, writeLock, clientSideKeypairStoreWriter, clientSideKeypairProvider, siteProvider, keysetManager, keypairGenerator, clock);
+            ClientSideKeypairService clientSideKeypairService = new ClientSideKeypairService(config, auth, writeLock, clientSideKeypairStoreWriter, clientSideKeypairProvider, siteProvider, keysetManager, keypairGenerator, clock);
```
atarassov-ttd, line 235: Curious why you removed `var` here and `val` below?

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
atarassov-ttd: You may also consider extracting this into a function.

src/main/java/com/uid2/admin/Main.java
sophia-chen-ttd: It was throwing errors on my side for some reason

src/main/java/com/uid2/admin/vertx/service/ServiceLinkService.java
```diff
@@ -107,11 +109,33 @@ private void handleServiceLinkAdd(RoutingContext rc) {
                 return;
             }
 
+            if (name.isEmpty()) {
```
gmsdelmundo, line 112: Maybe this can be combined with the first if-condition checking for nulls?

src/test/java/com/uid2/admin/vertx/ServiceServiceTest.java
```diff
@@ -418,6 +419,53 @@ void updateRoles(Vertx vertx, VertxTestContext testContext) {
         });
     }
 
+    @Test
+    void updateService_removeRolesNotUsedByServiceLinks_Succeeds(Vertx vertx, VertxTestContext testContext) {
```
gmsdelmundo, line 423: `Succeeds` -> `succeeds`

src/test/java/com/uid2/admin/vertx/ServiceServiceTest.java
```diff
@@ -418,6 +419,53 @@ void updateRoles(Vertx vertx, VertxTestContext testContext) {
...
+            assertEquals(200, response.statusCode());
+            existingService.setRoles(Set.of(Role.MAPPER, Role.SHARER));
+            checkServiceJson(existingService, response.bodyAsJsonObject());
+            verify(serviceStoreWriter, times(1)).upload(List.of(existingService), null);
+            testContext.completeNow();
+        });
+    }
+
+    @Test
+    void updateService_removeRolesInUseByServiceLinks_ReturnsError(Vertx vertx, VertxTestContext testContext) {
```
gmsdelmundo, line 447: `Returns` -> `returns`

src/main/java/com/uid2/admin/vertx/service/ServiceService.java
```diff
@@ -203,6 +205,13 @@ private void handleUpdate(RoutingContext rc) {
                     ResponseUtil.error(rc, 400, "invalid parameter: roles");
                     return;
                 }
+                // check that if role is removed, it is no longer in use by service links
+                Set<Role> rolesToRemove = service.getRoles().stream().filter(r -> !roles.contains(r)).collect(Collectors.toSet());
+                List<ServiceLink> serviceLinks = this.serviceLinkProvider.getAllServiceLinks().stream().filter(sl -> sl.getServiceId() == serviceId).collect(Collectors.toList());
+                if (serviceLinks.stream().anyMatch(sl -> !Collections.disjoint(sl.getRoles(), rolesToRemove))) {
```
vishalegbert-ttd, line 211: if (serviceLinks.stream().anyMatch(sl -> sl.getRoles().stream().anyMatch(rolesToRemove::contains)))

is simpler and more readable, avoid "not-not" logic as it's confusing

src/test/java/com/uid2/admin/vertx/ServiceLinkServiceTest.java
```diff
@@ -289,12 +442,64 @@ void updateNameSucceeds(Vertx vertx, VertxTestContext testContext) {
...
+        setServices(new Service(1, 123, "name1", Set.of(Role.CLIENTKEY_ISSUER, Role.MAPPER, Role.SHARER)));
+        ServiceLink existingLink = new ServiceLink("link1", 1, 123, "name1", Set.of(Role.MAPPER));
+        setServiceLinks(existingLink);
+
+        JsonObject jo = new JsonObject();
+        jo.put("link_id", "link1");
+        jo.put("service_id", 1);
+        jo.put("site_id", 123);
+        jo.put("name", "newname");
+        jo.put("roles", JsonArray.of(Role.ADMINISTRATOR));
```
asloobq, line 484: doesn't look like a valid use case to assign ADMINISTRATOR to a service link.

PR #184, opened 2023-12-05
Add get site by id endpoint

joshua-smee-ttd:
Added api call "/api/site/:siteId"

PR #191, opened 2023-12-15
Added end point to get keys by site id

joshua-smee-ttd:
Added endpoint to return all keys based on a siteId.

image
Also added search to call this endpoint on panel.

### DIFF
--- a/src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
@@ -76,7 +76,7 @@ public class ClientKeyService implements IService {
         router.get("/api/client/list").handler(
                 auth.handle(this::handleClientList, Role.CLIENTKEY_ISSUER));
 
-        router.get("/api/client/listBySite/:siteId").handler(
+        router.get("/api/client/list/:siteId").handler(
                 auth.handle(this::handleClientListBySite, Role.CLIENTKEY_ISSUER, Role.SHARING_PORTAL));
 
         router.get("/api/client/reveal").handler(
--- a/src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
@@ -72,8 +72,13 @@ public class ClientKeyService implements IService {
                 this.handleRewriteMetadata(ctx);
             }
         }, Role.CLIENTKEY_ISSUER));
+
         router.get("/api/client/list").handler(
                 auth.handle(this::handleClientList, Role.CLIENTKEY_ISSUER));
+
+        router.get("/api/client/listBySite/:siteId").handler(
+                auth.handle(this::handleClientListBySite, Role.CLIENTKEY_ISSUER));
+
         router.get("/api/client/reveal").handler(
                 auth.handle(this::handleClientReveal, Role.CLIENTKEY_ISSUER));
 
@@ -154,6 +159,36 @@ public class ClientKeyService implements IService {
                 jo.put("roles", RequestUtil.getRolesSpec(c.getRoles()));
                 jo.put("created", c.getCreated());
                 jo.put("site_id", c.getSiteId());
+                jo.put("disabled", c.isDisabled());
+                jo.put("service_id", c.getServiceId());
+            }
+
+            rc.response()
+                    .putHeader(HttpHeaders.CONTENT_TYPE, "application/json")
+                    .end(ja.encode());
+        } catch (Exception e) {
+            rc.fail(500, e);
+        }
+    }
+
+    private void handleClientListBySite(RoutingContext rc) {
+        try {
+            final Site site = RequestUtil.getSiteFromUrl(rc, "siteId", this.siteProvider);
+            if (site == null) {
+                return;
+            }
+
+            JsonArray ja = new JsonArray();
+            List<LegacyClientKey> collection = this.clientKeyProvider.getAll().stream().filter(legacyClientKey -> legacyClientKey.getSiteId() == site.getId()).collect(Collectors.toList());
+            for (LegacyClientKey c : collection) {
+                JsonObject jo = new JsonObject();
+                ja.add(jo);
+
+                jo.put("key_id", c.getKeyId());
+                jo.put("name", c.getName());
+                jo.put("contact", c.getContact());
+                jo.put("roles", RequestUtil.getRolesSpec(c.getRoles()));
+                jo.put("created", c.getCreated());
                 jo.put("disabled", c.isDisabled());
                 jo.put("service_id", c.getServiceId());
             }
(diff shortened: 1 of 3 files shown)


src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -72,8 +72,13 @@ public void setupRoutes(Router router) {
                 this.handleRewriteMetadata(ctx);
             }
         }, Role.CLIENTKEY_ISSUER));
+
         router.get("/api/client/list").handler(
                 auth.handle(this::handleClientList, Role.CLIENTKEY_ISSUER));
+
+        router.get("/api/client/listBySite/:siteId").handler(
+                auth.handle(this::handleClientListBySite, Role.CLIENTKEY_ISSUER));
```
alex-yau-ttd, line 80: I think we want `Role.SHARING_PORTAL` as well?

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -72,8 +72,13 @@ public void setupRoutes(Router router) {
                 this.handleRewriteMetadata(ctx);
             }
         }, Role.CLIENTKEY_ISSUER));
+
         router.get("/api/client/list").handler(
                 auth.handle(this::handleClientList, Role.CLIENTKEY_ISSUER));
+
+        router.get("/api/client/listBySite/:siteId").handler(
```
alex-yau-ttd, line 79: following the pattern in [SiteService.java](https://github.com/IABTechLab/uid2-admin/blob/86edd40783a9b3e5930d97917fce710619bbea1c/src/main/java/com/uid2/admin/vertx/service/SiteService.java#L64), perhaps just `/api/client/:siteId` would be better?

src/test/java/com/uid2/admin/vertx/ClientKeyServiceTest.java
```diff
@@ -235,6 +235,45 @@ public void clientRolesCreatesSiteKeyIfNoneExists(Set<Role> roles, boolean siteK
...
+
+        get(vertx, testContext, "api/client/listBySite/100", expectHttpStatus(testContext, 404));
+    }
+
+    @Test
+    public void listBySiteWorking(Vertx vertx, VertxTestContext testContext) {
+        fakeAuth(Role.CLIENTKEY_ISSUER);
+
+        setSites(new Site(666, "test2_site", true), new Site(999, "test_site", true));
+        setClientKeys(new LegacyClientBuilder().withSiteId(999).withName("999").build(), new LegacyClientBuilder().withSiteId(666).withName("666").build());
```
alex-yau-ttd, line 264: nit: I'm finding the code here a bit hard to read, could we possibly change to something like:
```
setSites(
    new Site(666, "test2_site", true),
    new Site(999, "test_site", true)
);
setClientKeys(
    new LegacyClientBuilder()
        .withSiteId(999)
        .withName("999")
        .build(),
    new LegacyClientBuilder()
        .withSiteId(666)
        .withName("666")
        .build()
);
```

src/test/java/com/uid2/admin/vertx/ClientKeyServiceTest.java
```diff
@@ -235,6 +235,45 @@ public void clientRolesCreatesSiteKeyIfNoneExists(Set<Role> roles, boolean siteK
...
+
+    @Test
+    public void listBySiteUnusedId(Vertx vertx, VertxTestContext testContext){
+        fakeAuth(Role.CLIENTKEY_ISSUER);
+
+        get(vertx, testContext, "api/client/listBySite/100", expectHttpStatus(testContext, 404));
+    }
+
+    @Test
+    public void listBySiteWorking(Vertx vertx, VertxTestContext testContext) {
```
alex-yau-ttd, line 260: I know there's some questions about naming conventions for our tests, but I think we can drop the `Working` at least: `listBySite`.

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -166,6 +171,36 @@ private void handleClientList(RoutingContext rc) {
...
+                return;
+            }
+
+            JsonArray ja = new JsonArray();
+            List<LegacyClientKey> collection = this.clientKeyProvider.getAll().stream().filter(legacyClientKey -> legacyClientKey.getSiteId() == site.getId()).collect(Collectors.toList());
+            for (LegacyClientKey c : collection) {
+                JsonObject jo = new JsonObject();
+                ja.add(jo);
+
+                jo.put("key_id", c.getKeyId());
```
alex-yau-ttd, line 187: What do you think about giving this a similar refactoring to what you did for the [SiteService](https://github.com/IABTechLab/uid2-admin/pull/184/files#diff-f6fd75edc9be9de7cd9f625255750e10d7b1d2df3093eafe7fc042bbe4dc8b8f)? That way `handleClientList` and `handleClientListBySite` don't have so much duplicated code.

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
joshua-smee-ttd: I think that route may become misleading. I would think it would possibly more look like that endpoint would provide a single key based on the id?

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
joshua-smee-ttd: Ah yeah we would sorry.

src/test/java/com/uid2/admin/vertx/ClientKeyServiceTest.java
joshua-smee-ttd: Okay cool. Will do that.

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
joshua-smee-ttd: I think either way is valid. I left it this way because I felt it would encourage removing data from the object more if there's data we don't want to pass to the portal and we can't quite as accidentally pass new data.

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
alex-yau-ttd: what about just `/api/client/list/:siteId`? That pattern is also used in `SharingService.java`

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
alex-yau-ttd: alrighty - we can always refactor later if we change our minds.

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
joshua-smee-ttd: Sounds good. Have made that change.

alex-yau-ttd (approved):

PR #199, opened 2024-01-02
Separating Contact and Name field for client keys

joshua-smee-ttd:
Separating contact and name fields of clients keys. This way contact can be used for logging and is uniquely identifiably for TTD tracking while the name can be anything, allowing users to give keys names relevant to them. Names do not have to unique anymore. Contact will still be unique and is generated based on a combination of the site name and the keyId.

chrome_BjxchURblE
Changed layout of page for more logical flow, now each operation will have popups for the different information they need apart from contact.

image
image
Changing the contact of a key will require the new contact to be filled in in a new popup along with a second confirmation.

### DIFF
--- a/src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
@@ -486,16 +486,16 @@ public class ClientKeyService implements IService {
 
             List<LegacyClientKey> clients = getAllClientKeys();
 
-            LegacyClientKey c = existingClient.get();
-            c.withContact(newContact);
+            LegacyClientKey existingClientObject = existingClient.get();
+            existingClientObject.withContact(newContact);
 
             // upload to storage
             storeWriter.upload(clients, null);
 
-            this.keysetManager.createKeysetForClient(c.toClientKey());
+            this.keysetManager.createKeysetForClient(existingClientObject.toClientKey());
 
             // return client with new key
-            rc.response().end(JSON_WRITER.writeValueAsString(c.toClientKey()));
+            rc.response().end(JSON_WRITER.writeValueAsString(existingClientObject.toClientKey()));
         } catch (Exception e) {
             rc.fail(500, e);
         }
--- a/src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
@@ -118,6 +118,12 @@ public class ClientKeyService implements IService {
             }
         }, Role.CLIENTKEY_ISSUER));
 
+        router.post("/api/client/contact").blockingHandler(auth.handle((ctx) -> {
+            synchronized (writeLock) {
+                this.handleClientContact(ctx);
+            }
+        }, Role.CLIENTKEY_ISSUER));
+
         router.post("/api/client/rename").blockingHandler(auth.handle((ctx) -> {
             synchronized (writeLock) {
                 this.handleClientRename(ctx);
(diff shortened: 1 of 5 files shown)


alex-yau-ttd:
I think parts of the UI are still unintuitive to me. For instance, the first three inputs under `Existing Key` only require `Contact` while the next two require `Roles` or `Site Id` to be populated and the last one (`Rename Client Key`) only requires a `Contact` but then opens a dialog to input the new contact name. 

e.g. Updating a Site requires these inputs
image.

I think consistency would be good here, we could either:
1. Change the form so that each button has its own inputs, e.g. `Reveal Client Key` has its own `Contact` field, so does `Disable Client Key`, etc., or
2. Change the `Set Roles` and `Update Site` buttons to also use dialogs.

Also open to other ideas - what are your thoughts?

webroot/adm/client-key.html
```diff
@@ -159,15 +168,26 @@ <h3>Output</h3>
...
-            var oldClientName = encodeURIComponent($('#clientName').val());
-            var url = '/api/client/rename?oldName=' + oldClientName + '&newName=' + newClientName;
+            var clientContact = encodeURIComponent($('#clientContact').val());
+            var newClientName = encodeURIComponent($('#clientName').val());
+            var url = '/api/client/rename?contact=' + clientContact + '&newName=' + newClientName;
+            doApiCall('POST', url, '#standardOutput', '#errorOutput');
+        });
+
+        $('#doSetContact').on('click', function () {
+            var newClientContact = prompt("Enter the new contact", "new contact here");
```
alex-yau-ttd, line 178: I'm not sure if we need the placeholder here, I think it might make it more likely to accidentally rename a client to `new contact here`. Could we possibly use `confirm` and just have it fail when there is no new `contact` provided?

alex-yau-ttd: Also hitting `Cancel` on the `prompt` results in:
image

joshua-smee-ttd:
I think the dialogs makes more sense to me. I'll put those in and see what it looks like.

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -461,29 +460,63 @@ private void handleClientRoles(RoutingContext rc) {
...
+                    .stream().filter(c -> c.getContact().equals(newContact))
+                    .findFirst();
+            if (existingClientWithNewContact.isPresent()) {
+                ResponseUtil.error(rc, 400, "Client with contact already exists. Contact must be unique.");
+                return;
+            }
+
+            List<LegacyClientKey> clients = getAllClientKeys();
+
+            LegacyClientKey c = existingClient.get();
```
alex-yau-ttd, line 489: nit: can we please give this a better name? `c` doesn't mean much

joshua-smee-ttd:
image
Have updated to look like this.  Also updated description.

alex-yau-ttd:
> image Have updated to look like this. Also updated description.

nice! what are your thoughts on changing `New Key` to `Add Key` and `Existing Key` to `Edit Key`? Perhaps `Add Client Key` could be changed to a button since it is not part of a list of operations now?

src/test/java/com/uid2/admin/vertx/ClientKeyServiceTest.java
```diff
@@ -279,16 +268,48 @@ public void listBySite(Vertx vertx, VertxTestContext testContext) {
...
+            ClientKey expected = new LegacyClientBuilder().withContact("test_contact1").build().toClientKey();
+            assertAll(
+                    "clientSetContact",
+                    () -> assertEquals(200, response.statusCode()),
+                    () -> assertEquals(expected, OBJECT_MAPPER.readValue(response.bodyAsString(), ClientKey.class)),
+                    () -> verify(clientKeyStoreWriter).upload(collectionOfSize(1), isNull())
+            );
+            testContext.completeNow();
+        });
+
```
alex-yau-ttd, line 287: nit: there's an unnecessary line here

src/test/java/com/uid2/admin/vertx/ClientKeyServiceTest.java
```diff
@@ -279,16 +268,48 @@ public void listBySite(Vertx vertx, VertxTestContext testContext) {
...
+    @Test
+    public void setContactWithExistingContact(Vertx vertx, VertxTestContext testContext) {
+        fakeAuth(Role.CLIENTKEY_ISSUER);
+
+        setClientKeys(
+                new LegacyClientBuilder().build(),
+                new LegacyClientBuilder().withContact("test_contact1").build()
+        );
+
```
alex-yau-ttd, line 298: nit: there's an unnecessary line here

joshua-smee-ttd:
image
Yeah that sounds good. Have made those changes.

webroot/adm/client-key.html
```diff
@@ -105,69 +102,105 @@ <h3>Output</h3>
...
-            var serviceId = encodeURIComponent($('#serviceId').val());
-            var url = '/api/client/update?name=' + clientName + '&site_id=' + siteId + '&service_id=' + serviceId;
+            var clientContact = encodeURIComponent($('#clientContact').val());
+
+            var siteId = prompt("Enter the new siteId", "");
+            if (!siteId){
+                if (siteId === '') $('#errorOutput').html("Required Parameter: siteId");
+                return;
+            }
+            var serviceId = prompt("Enter the new serviceId", "");
```
alex-yau-ttd, line 132: Do you know if there is some link between `serviceId` and `siteId`? To me, they are distinct, and probably should be separate buttons.

src/test/java/com/uid2/admin/vertx/ClientKeyServiceTest.java
```diff
@@ -279,16 +268,46 @@ public void listBySite(Vertx vertx, VertxTestContext testContext) {
...
+                    () -> assertEquals(200, response.statusCode()),
+                    () -> assertEquals(expected, OBJECT_MAPPER.readValue(response.bodyAsString(), ClientKey.class)),
+                    () -> verify(clientKeyStoreWriter).upload(collectionOfSize(1), isNull())
+            );
+            testContext.completeNow();
+        });
+    }
+
+    @Test
+    public void setContactWithExistingContact(Vertx vertx, VertxTestContext testContext) {
```
alex-yau-ttd, line 290: Is it worth testing the 404 case in a separate test too? `setContactWithUnknownContact`?

webroot/adm/client-key.html
```diff
@@ -12,51 +12,48 @@ <h1>UID2 Admin - Client Key Management</h1>
...
-<label for="serviceId">Service Id:</label>
-<input type="text" id="serviceId" name="serviceId" value="0">
-<br>
-<br>
-
-<h3>Operations</h3>
-
+<h3>Add Key</h3>
+<form id="addKeyForm">
+<label for="addClientName">Name:</label>
```
alex-yau-ttd, line 29: Let's fix up the indentation of everything within the `<form>`.

webroot/adm/client-key.html
joshua-smee-ttd: The update site was utilizing this endpoint already which updated both. I believe that serviceId is related to the siteId in some manner however so should be changed.

alex-yau-ttd (approved):

PR #202, opened 2024-01-10
Checks for empty name and contact in client key

joshua-smee-ttd:


src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -508,6 +516,11 @@ private void handleClientRename(RoutingContext rc) {
 
             final String contact = rc.queryParam("contact").get(0);
             final String newName = rc.queryParam("newName").get(0);
+            if (newName.isEmpty()){
```
cYKatherine, line 519: should we check for `contact` being empty?

cYKatherine:
Can we also add the old client key name here: https://github.com/IABTechLab/uid2-admin/blob/main/webroot/adm/client-key.html#L177

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
joshua-smee-ttd: I'm in two minds about this. I could add a check to every endpoint but now there is no way for a contact to end up equal to "". I also prefer this method because if somehow a `contact` does end up as "" again since this way we can actually change it. Functionally it just throws an error when it can't find the contact.

joshua-smee-ttd:
> Can we also add the old client key name here: https://github.com/IABTechLab/uid2-admin/blob/main/webroot/adm/client-key.html#L177
This is not particularly easy to do since when the call is made the front end has no knowledge of the old name. I could get around this by making a call to admin to get the name first but this feels unnecessary? Happy to add it if you think it's particularly important.

cYKatherine:
> > Can we also add the old client key name here: https://github.com/IABTechLab/uid2-admin/blob/main/webroot/adm/client-key.html#L177
> > This is not particularly easy to do since when the call is made the front end has no knowledge of the old name. I could get around this by making a call to admin to get the name first but this feels unnecessary? Happy to add it if you think it's particularly important.

Maybe change the wording to `Are you sure you want to rename the key with contact " + clientContact " to: " + newClientName`?

cYKatherine (approved):

PR #205, opened 2024-01-15
Allow search by keyId and allow portal to use edit endpoints

joshua-smee-ttd:
Added a new endpoint to allow to search for client key by siteId. Added button to ui.
image

Also allowed portal to use edit endpoints to allow it to edit apiKeys.

### DIFF
--- a/src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
@@ -217,7 +217,6 @@ public class ClientKeyService implements IService {
         }
     }
 
-
     private static JsonObject getClientReturnObject(LegacyClientKey clientKey) {
         JsonObject returnObject = new JsonObject();
 
--- a/src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
@@ -6,27 +6,30 @@ import com.uid2.admin.legacy.LegacyClientKey;
 import com.uid2.admin.legacy.LegacyClientKeyStoreWriter;
 import com.uid2.admin.legacy.RotatingLegacyClientKeyProvider;
 import com.uid2.admin.managers.KeysetManager;
-import com.uid2.shared.model.Site;
-import com.uid2.shared.secret.IKeyGenerator;
 import com.uid2.admin.vertx.JsonUtil;
 import com.uid2.admin.vertx.RequestUtil;
 import com.uid2.admin.vertx.ResponseUtil;
 import com.uid2.admin.vertx.WriteLock;
 import com.uid2.shared.auth.Role;
 import com.uid2.shared.middleware.AuthMiddleware;
+import com.uid2.shared.model.Site;
+import com.uid2.shared.secret.IKeyGenerator;
 import com.uid2.shared.secret.KeyHashResult;
 import com.uid2.shared.secret.KeyHasher;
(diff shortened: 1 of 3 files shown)


src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -208,6 +191,46 @@ private void handleClientListBySite(RoutingContext rc) {
...
+
+            rc.response()
+                    .putHeader(HttpHeaders.CONTENT_TYPE, "application/json")
+                    .end(getClientReturnObject(clientKey).encode());
+        } catch (Exception e) {
+            rc.fail(500, e);
+        }
+    }
+
+    private static JsonObject getClientReturnObject(LegacyClientKey clientKey) {
```
alex-yau-ttd, line 220: when we did this for `SiteService.java`, we called it `createSiteJsonObject`. Perhaps we could call this one `createClientKeyJsonObject` or similar?

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -208,6 +191,46 @@ private void handleClientListBySite(RoutingContext rc) {
         }
     }
 
+    private void handleClientByKeyId(RoutingContext rc) {
+        try {
+            String keyId;
+
+            try{
```
alex-yau-ttd, line 198: I wonder if there is a more elegant approach to this that avoids having to nest `try-catch`es. What about something similar to the `clientKey` below like:
```
String keyId = rc.queryParam("keyId").stream().findFirst().orElse(null);

if (keyId == null) {
    ResponseUtil.error(rc, 400, "unable to parse keyId ");
    return;
}
```

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
```diff
@@ -208,6 +191,46 @@ private void handleClientListBySite(RoutingContext rc) {
...
+                keyId = rc.queryParam("keyId").get(0);
+            } catch (Exception e) {
+                ResponseUtil.error(rc, 400, "unable to parse keyId " + e.getMessage());
+                return;
+            }
+
+            LegacyClientKey clientKey = this.clientKeyProvider.getAll().stream().filter(legacyClientKey -> legacyClientKey.getKeyId().equals(keyId)).findFirst().orElse(null);
```
alex-yau-ttd, line 205: I know it's only a one-liner, but it's a pretty long line. I wonder if it is worth extracting to:
```
private LegacyClientKey getClientKey(String keyId) {
    return this.clientKeyProvider.getAll()
            .stream()
            .filter(legacyClientKey -> legacyClientKey.getKeyId().equals(keyId))
            .findFirst()
            .orElse(null);
}
```

alex-yau-ttd: Means we can better separate concerns - all I care about at this level of abstraction is that I'm getting a client key based off the `keyId`

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
joshua-smee-ttd: Ah that's much nicer!

src/main/java/com/uid2/admin/vertx/service/ClientKeyService.java
joshua-smee-ttd: Yeah that makes sense!

alex-yau-ttd (approved):

PR #210, opened 2024-01-24
sch-UID2-2516-redirect-snowflake-etl-to-read-from-service-links-implementation

sophia-chen-ttd:
We'll be using uid2-admin to host the metrics for the snowflake account metadata and number of snowflake accounts. I've added a new data store method to retrieve the number of snowflake service links, and added a new data store for service link version.

I've also added the ID_READER and GENERATOR roles to the local snowflake service to reflect the roles required for the snowflake accounts.

### DIFF
--- a/src/main/java/com/uid2/admin/Main.java
+++ b/src/main/java/com/uid2/admin/Main.java
@@ -310,7 +310,7 @@ public class Main {
             DataStoreMetrics.addDataStoreMetrics("salt", saltProvider);
             DataStoreMetrics.addDataStoreMetrics("partners", partnerConfigProvider);
             DataStoreMetrics.addDataStoreMetrics("service_link", serviceLinkProvider);
-            DataStoreMetrics.addDataStoreSnowflakeMetrics("snowflake_service_link", serviceLinkProvider, serviceProvider);
+            DataStoreMetrics.addDataStoreServiceLinkEntryCount("snowflake", serviceLinkProvider, serviceProvider);
 
 
             ReplaceSharingTypesWithSitesJob replaceSharingTypesWithSitesJob = new ReplaceSharingTypesWithSitesJob(config, writeLock, adminKeysetProvider, keysetProvider, keysetStoreWriter, siteProvider);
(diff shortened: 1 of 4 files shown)


src/main/java/com/uid2/admin/monitoring/DataStoreMetrics.java
```diff
@@ -21,4 +26,23 @@ public static void addDataStoreMetrics(String dataType, IMetadataVersionedStore
                 .description("version from metadata of a data store")
                 .register(globalRegistry);
     }
+    public static void addDataStoreSnowflakeMetrics(String dataType, RotatingServiceLinkStore serviceLinkStore, RotatingServiceStore serviceStore) {
+        Gauge
+                .builder("uid2_data_store_entry_count", () -> {
+                    try {
+                        // Warning: this downloads metadata from the underlying remote data store
```
mcollins-ttd, line 33: I thought it was the call to `getMetadata` in the method above that downloads metadata. Where are we downloading metadata in this method?

sophia-chen-ttd: That's true, I haven't loaded the content yet. I'll fix that!

src/main/java/com/uid2/admin/monitoring/DataStoreMetrics.java
```diff
@@ -21,4 +26,25 @@ public static void addDataStoreMetrics(String dataType, IMetadataVersionedStore
...
+                        serviceStore.loadContent(serviceStore.getMetadata());
+                        serviceLinkStore.loadContent(serviceLinkStore.getMetadata());
+                        Optional<Service> snowflakeService = serviceStore.getAllServices().stream().filter(s -> s.getName().equals("snowflake")).findFirst();
+                        if (snowflakeService.isEmpty()) { throw new IllegalStateException("snowflake service does not exist, unable to find snowflake accounts"); }
```
atarassov-ttd, line 37: EUID does not have snowflake.

src/main/java/com/uid2/admin/Main.java
```diff
@@ -309,6 +309,9 @@ public void run() {
             DataStoreMetrics.addDataStoreMetrics("enclaves", enclaveIdProvider);
             DataStoreMetrics.addDataStoreMetrics("salt", saltProvider);
             DataStoreMetrics.addDataStoreMetrics("partners", partnerConfigProvider);
+            DataStoreMetrics.addDataStoreMetrics("service_link", serviceLinkProvider);
+            DataStoreMetrics.addDataStoreServiceLinkEntryCount("snowflake", serviceLinkProvider, serviceProvider);
```
vishalegbert-ttd, line 313: I think we should make this method specific to Snowflake, I don't see how it could be used for other service links

Therefore we don't need to pass in the service name, just hardcode it (with a constant) inside the method and make the method name Snowflake specific

PR #211, opened 2024-01-25
Introducing swagger in the Admin API [Hackathon project] 

asloobq:
To access -> 
1] run the app and visit http://localhost:8089/swagger-ui/#/
2] Get your API Token. You can do this by calling GET /api/token/get and copying the key. Or just open admin UI, select any resource which will call the endpoint in the background and copy the token from the network tab
3] Click 'Authorize' (top right) and paste the value. 

Benefits

- No need to guess what operations require which fields
- Know exactly which endpoints are available. No second guessing
- Easier to expose API for usage than current UI by just defining the OpenAPI spec (which could be part of the design doc)

Pending stuff

- Other resources like operator keys, keysets, etc.
- Automatic API token retrieval
- Role requirements for each endpoint
Swagger-admin-screencap


asloobq:
Closing as not needed since we have a new Admin UI https://github.com/IABTechLab/uid2-admin/pull/524

PR #221, opened 2024-02-21
Ian UID2-2690 guard endpoints with okta

Ian-Nara:
There are some TODOs in there that indicate where changes will be needed once we are ready to implement role granularity


src/main/java/com/uid2/admin/auth/OktaAuthProvider.java
```diff
@@ -11,16 +12,34 @@
...
 import java.time.Duration;
-import java.util.Arrays;
 import java.util.List;
-import java.util.Collections;
 
 public class OktaAuthProvider implements AuthProvider {
     private final JsonObject config;
-    private List<String> scopes = Collections.unmodifiableList(Arrays.asList("openid",  "profile" ,"email" ,"groups"));
-
+    private final List<String> scopes = List.of("openid", "profile", "email", "uid2.admin.human", "offline_access");
```
asloobq, line 19: do we still need these scopes ? I don't see them referenced anywhere

asloobq (commented):
reviewed partially. will continue tomorrow

src/main/java/com/uid2/admin/auth/OktaAuthProvider.java
```diff
@@ -11,16 +12,34 @@
...
+    private final IdTokenVerifier idTokenVerifier;
     public OktaAuthProvider(JsonObject config) {
         this.config = config;
+        if(isAuthDisabled(config)) {
+            this.accessTokenVerifier = null;
+            this.idTokenVerifier = null;
+            return;
+        }
+        this.accessTokenVerifier = JwtVerifiers.accessTokenVerifierBuilder()
+                .setIssuer(config.getString("okta_auth_server"))
```
asloobq, line 30: might be better to declare string const for all okta config as they are being used in multiple places

src/main/java/com/uid2/admin/auth/TokenRefreshHandler.java
```diff
@@ -0,0 +1,75 @@
...
+            refreshToken = rc.user().principal().getString("refresh_token");
+        }
+        if (idToken != null && refreshToken != null) {
+            try {
+                idTokenVerifier.decode(idToken, null);
+            } catch (JwtVerificationException e) {
+                refreshToken(rc, refreshToken);
+            }
+        }
+        rc.next();
```
asloobq, line 58: you are not checking whether Access Token is present and expired before moving on ?

src/main/java/com/uid2/admin/auth/TokenRefreshHandler.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private void refreshToken(RoutingContext rc, String refreshToken) {
+        HttpResponse<String> response;
+        try {
+            response = httpClient.post(String.format("%s/v1/token?grant_type=refresh_token&refresh_token=%s", this.authServer, refreshToken), "", this.authHeaders);
+        } catch (IOException e) {
+            return;
```
asloobq, line 66: does vertx know that auth failed and will reply with 401 ? or do you need to set the status code ?

src/main/java/com/uid2/admin/vertx/AdminVerticle.java
```diff
@@ -57,19 +59,20 @@ public void start(Promise<Void> startPromise) {
 
     private Router createRoutesSetup() {
         final Router router = Router.router(vertx);
-        router.route().handler(SessionHandler.create(LocalSessionStore.create(vertx)));
+        router.route().handler(SessionHandler.create(LocalSessionStore.create(vertx)).setSessionTimeout(32400000)); // 9 hr session timeout
```
asloobq, line 62: may be better to make it a config setting to have different timeout by environment ?

src/main/java/com/uid2/admin/vertx/AdminVerticle.java
```diff
@@ -57,19 +59,20 @@ public void start(Promise<Void> startPromise) {
...
         final AuthenticationHandler oktaHandler = this.authProvider.createAuthHandler(vertx, router.route("/oauth2-callback"));
+
         router.route().handler(BodyHandler.create());
         router.route().handler(StaticHandler.create("webroot"));
 
         router.route("/login").handler(oktaHandler);
-        router.route("/adm/*").handler(oktaHandler); 
+        router.route("/adm/*").handler(oktaHandler);
```
asloobq, line 69: do we still have routes with /adm/* ? Or do they go away when you delete AdminKeyService.java ?

src/main/java/com/uid2/admin/auth/AdminAuthMiddleware.java
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+
+        private void validateIdToken(RoutingContext rc, String idToken) {
+            Jwt jwt;
+            try {
+                jwt = authProvider.getIdTokenVerifier().decode(idToken, null);
+            } catch (JwtVerificationException e) {
+                rc.session().destroy();
+                rc.response().putHeader("REQUIRES_AUTH", "1").setStatusCode(401).end();
```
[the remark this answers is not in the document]
Ian-Nara: It is because, for machines using the client_credentials workflow, we always begin the authentication process from the start (no refresh_tokens). For the UI however, we use this to indicate that reauthentication is needed via /login, in the event the refresh_token is expired.

src/main/java/com/uid2/admin/auth/OktaAuthProvider.java
Ian-Nara: just openid, email, and uid2.admin.human (openid is required by default, email to display the user logged in in the UI)

src/main/java/com/uid2/admin/auth/TokenRefreshHandler.java
Ian-Nara: Token refresh is available for the human user workflow only. We only care about the id_token / the groups.

src/main/java/com/uid2/admin/vertx/AdminVerticle.java
Ian-Nara: /adm/* matches against UI interactions by webroot/adm/*.html

src/main/java/com/uid2/admin/auth/TokenRefreshHandler.java
Ian-Nara: Hmm it will not, and we could reject the request at this point. I was thinking it would be better to let full responsibility for authorization-based rejections be contained in AdminAuthMiddleware though. I think it makes it more clear where in the code the failure is occurring.

conf/local-config.json
```diff
@@ -32,10 +32,12 @@
...
   "salt_snapshot_activates_in_seconds": 180,
   "salt_snapshot_expires_after_seconds": 180000,
-  "is_auth_disabled": true,
+  "is_auth_disabled": false,
   "enable_keysets": true,
   "okta_client_id": "0oa1ok419hd6kbKJZ1d8",
   "okta_client_secret": "GET_FROM_1PASSWORD",
-  "okta_issuer": "https://uid2.okta.com/",
-  "okta_callback": "http://localhost:8089/oauth2-callback"
+  "okta_auth_server": "https://uid2.okta.com/oauth2/aus1oqu660mF7W3hi1d8",
```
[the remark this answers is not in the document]
Ian-Nara: Yeah for local development that doesn't involve auth, setting is_auth_disabled is best. Updated this back to true in local-config.json

webroot/js/main.js
```diff
@@ -8,15 +8,25 @@ function doApiCall(method, url, outputDiv, errorDiv, body) {
...
-            "Authorization": authHeader
-        },
         data : body,
         success: function (text) {
             var pretty = JSON.stringify(JSON.parse(text),null,2);
             $(outputDiv).text(pretty);
         },
-        error: function (err) { standardErrorCallback(err, errorDiv) }
+        error: function (err, status, header) {
+            if(err.getResponseHeader("REQUIRES_AUTH") == 1) {
```
asloobq, line 17: this code is repeated below, probably better to refactor it to a method like `standardErrorCallback` is and reuse that ?

webroot/js/main.js
```diff
@@ -63,46 +83,51 @@ function doApiCallWithCallback(method, url, onSuccess, onError, body) {
...
-        url: '/api/token/get',
+        url: '/api/userinfo',
         dataType: 'text',
         success: function (text) {
             var u = JSON.parse(text);
-            $('#loginEmail').text(u.contact);
+            $('#loginEmail').text(u.email);
             $('.authed').show();
-            if (u.roles.findIndex(e => e === 'CLIENTKEY_ISSUER') >= 0) {
+            if (u.groups.findIndex(e => e === 'developer' || e === 'developer-elevated' || e === 'infra-admin' || e === 'admin') >= 0) {
```
asloobq, line 116: all these if conditions below can be merged into this.

asloobq (commented):
completed review. some refactoring needed in UI JS code. Otherwise looks good

src/main/java/com/uid2/admin/vertx/AdminVerticle.java
```diff
@@ -57,19 +59,20 @@ public void start(Promise<Void> startPromise) {
...
+        router.route("/api/*").handler(tokenRefreshHandler);
 
         router.get("/login").handler(new RedirectToRootHandler(false));
         router.get("/logout").handler(new RedirectToRootHandler(true));
         router.get("/ops/healthcheck").handler(this::handleHealthCheck);
-        router.get("/api/token/get").handler(ctx -> handleTokenGet(ctx));
-        router.get("/protected").handler(this::handleProtected);
+        router.get("/api/userinfo").handler(this::handleUserinfo);
```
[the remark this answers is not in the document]
asloobq: realized its being used by Admin UI

src/main/java/com/uid2/admin/vertx/AdminVerticle.java
```diff
@@ -86,56 +89,20 @@ private void handleHealthCheck(RoutingContext rc) {
...
-    }
-
-    String getEmailClaim(RoutingContext ctx) {
+    private void handleUserinfo(RoutingContext rc) {
+        if(isAuthDisabled(config)) rc.response().setStatusCode(200).end(
+                JsonObject.of("groups", JsonArray.of("developer", "developer-elevated", "infra-admin", "admin"), "email", "test.user@unifiedid.com").toString());
         try {
-            Jwt jwt = this.authProvider.createTokenVerifier().decode(ctx.user().principal().getString("id_token"));
-            return jwt.getClaims().get("email").toString();
+            Jwt idJwt = this.authProvider.getIdTokenVerifier().decode(rc.user().principal().getString("id_token"), null);
```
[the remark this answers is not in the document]
Ian-Nara: It is not necessary as Exception will catch NPE. Only the UI needs to use this endpoint and if there is not a user then the response is the same, we need to authenticate. I will add a check for rc.session() == null though.

conf/local-config.json
```diff
@@ -32,10 +32,13 @@
   "site_key_expires_after_seconds": 90000,
   "salt_snapshot_activates_in_seconds": 180,
   "salt_snapshot_expires_after_seconds": 180000,
-  "is_auth_disabled": true,
+  "is_auth_disabled": false,
```
asloobq, line 35: enable auth again ?

asloobq (approved):
LGTM 👍

asloobq (approved):

PR #237, opened 2024-03-06
UID2-2922 update shared action names

cYKatherine:


thomasm-ttd (approved):

.trivyignore
```diff
@@ -0,0 +1,6 @@
+# List any vulnerability that are to be accepted
+# See https://aquasecurity.github.io/trivy/v0.35/docs/vulnerability/examples/filter/ 
+# for more details
+# e.g. 
+# CVE-2022-3996
+CVE-2023-52425
```
gmsdelmundo, line 6: Let's add the Jira ticket number here? So we know it's actually being tracked

gmsdelmundo (approved):

PR #270, opened 2024-04-25
Adding the ability to rename sites

cody-constine-ttd:
Before Rename:
image
After Rename:
image

### DIFF
--- a/src/main/java/com/uid2/admin/vertx/service/SiteService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/SiteService.java
@@ -1,6 +1,5 @@
 package com.uid2.admin.vertx.service;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectWriter;
 import com.uid2.admin.auth.AdminAuthMiddleware;
 import com.uid2.admin.legacy.ILegacyClientKeyProvider;
@@ -183,18 +182,7 @@ public class SiteService implements IService {
             siteProvider.loadContent();
 
             final String name = rc.queryParam("name").isEmpty() ? "" : rc.queryParam("name").get(0).trim();
-            if (name == null || name.isEmpty()) {
-                ResponseUtil.error(rc, 400, "must specify a valid site name");
-                return;
-            }
-
-            Optional<Site> existingSite = this.siteProvider.getAllSites()
-                    .stream().filter(c -> c.getName().equals(name))
-                    .findFirst();
-            if (existingSite.isPresent()) {
-                ResponseUtil.error(rc, 400, "site existed");
-                return;
-            }
+            if (validateSiteName(rc, name)) return;
 
             List<String> normalizedDomainNames = new ArrayList<>();
 
@@ -389,7 +377,8 @@ public class SiteService implements IService {
                     ResponseUtil.error(rc, 400, "Invalid parameter for visible: " + visibleParam);
                 }
             }
-            if(name != null) {
+            if (name != null) {
+                if (validateSiteName(rc, name)) return;
                 existingSite.setName(name);
             }
 
@@ -397,6 +386,22 @@ public class SiteService implements IService {
         } catch (Exception e) {
             rc.fail(500, e);
         }
+    }
+
+    private boolean validateSiteName(RoutingContext rc, String name) {
+        if (name == null || name.isEmpty()) {
+            ResponseUtil.error(rc, 400, "must specify a valid site name");
+            return true;
+        }
+
+        Optional<Site> existingSite = this.siteProvider.getAllSites()
+                .stream().filter(c -> c.getName().equals(name))
+                .findFirst();
+        if (existingSite.isPresent()) {
+            ResponseUtil.error(rc, 400, "site existed");
+            return true;
+        }
+        return false;
     }
 
     private static List<String> getNormalizedDomainNames(RoutingContext rc, JsonArray domainNamesJa) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/uid2/admin/vertx/service/SiteService.java
```diff
@@ -388,6 +389,9 @@ private void handleSiteUpdate(RoutingContext rc) {
                     ResponseUtil.error(rc, 400, "Invalid parameter for visible: " + visibleParam);
                 }
             }
+            if(name != null) {
```
cYKatherine, line 392: Do we check if the new name is duplicated to the existing sites?

src/test/java/com/uid2/admin/vertx/SiteServiceTest.java
```diff
@@ -876,4 +876,25 @@ void addSiteWithDuplicateAppNames(Vertx vertx, VertxTestContext testContext) {
             testContext.completeNow();
         });
     }
+
+    @Test
```
cYKatherine, line 880: Maybe add a test case when the newName is duplicated to the existing site names

src/main/java/com/uid2/admin/vertx/service/SiteService.java
cYKatherine: Also maybe a sanity check for empty string `name != ''`, we had a bug before where name was set to empty string and caused some hassle debugging :(

thomasm-ttd, line 392: Spacing around the if

src/test/java/com/uid2/admin/vertx/SiteServiceTest.java
```diff
@@ -876,4 +876,25 @@ void addSiteWithDuplicateAppNames(Vertx vertx, VertxTestContext testContext) {
...
+        setSites(sites);
+
+        post(vertx, testContext, "api/site/update?id=11&name=NewName", null, response -> {
+            assertEquals(200, response.statusCode());
+            assertEquals("NewName", response.bodyAsJsonObject().getString("name"));
+            assertEquals("NewName", sites[0].getName());
+            testContext.completeNow();
+        });
+    }
 }
```
thomasm-ttd, line 900: How about testing that the update hasn't affected other sites? I know the code currently doesn't have that issue, but is it worth testing that?

src/main/java/com/uid2/admin/vertx/service/SiteService.java
```diff
@@ -388,13 +377,33 @@ private void handleSiteUpdate(RoutingContext rc) {
...
+                existingSite.setName(name);
+            }
 
             uploadSiteToStoreWriterAndWriteExistingSiteToResponse(existingSite, rc);
         } catch (Exception e) {
             rc.fail(500, e);
         }
     }
 
+    private boolean validateSiteName(RoutingContext rc, String name) {
```
thomasm-ttd, line 391: This feels like the wrong way round to me - I would expect something called validateSiteName to return true if is was valid, and false is it was not.

src/main/java/com/uid2/admin/vertx/service/SiteService.java
```diff
@@ -388,13 +377,33 @@ private void handleSiteUpdate(RoutingContext rc) {
                     ResponseUtil.error(rc, 400, "Invalid parameter for visible: " + visibleParam);
                 }
             }
+            if (name != null) {
+                if (validateSiteName(rc, name)) return;
```
thomasm-ttd, line 381: It appears the convention in this file (which I also prefer) is that if statements have a following brace: 
if (validateSitenName(rc, name)) {
    return;
}

src/main/java/com/uid2/admin/vertx/service/SiteService.java
```diff
@@ -388,13 +377,33 @@ private void handleSiteUpdate(RoutingContext rc) {
...
+        if (name == null || name.isEmpty()) {
+            ResponseUtil.error(rc, 400, "must specify a valid site name");
+            return true;
+        }
+
+        Optional<Site> existingSite = this.siteProvider.getAllSites()
+                .stream().filter(c -> c.getName().equals(name))
+                .findFirst();
+        if (existingSite.isPresent()) {
+            ResponseUtil.error(rc, 400, "site existed");
```
alex-yau-ttd, line 401: nit: could we possibly align on some language and do something like we do in [ServiceService.java](https://github.com/IABTechLab/uid2-admin/blob/main/src/main/java/com/uid2/admin/vertx/service/ServiceService.java#L135)

So:
```
ResponseUtil.error(rc, 400, "site with name" + name + " already exists");
```

src/test/java/com/uid2/admin/vertx/SiteServiceTest.java
```diff
@@ -876,4 +876,63 @@ void addSiteWithDuplicateAppNames(Vertx vertx, VertxTestContext testContext) {
...
+
+        Site[] sites = {
+                new Site(11, "site1", false),
+                new Site(12, "site2", true),
+                new Site(13, "site3", false, Set.of("test1.com", "test2.net")),
+                new Site(14, "site4", false, null, Set.of("test1.com", "test2.net"), Set.of("com.123.game.app.android", "12345678")),
+        };
+        setSites(sites);
+
+        post(vertx, testContext, "api/site/update?id=11&name=site2", null, response -> {
```
alex-yau-ttd, line 916: nit: I think extracting this case to a separate test `renameSiteTestDuplicateName` would help with readability

thomasm-ttd (approved):

alex-yau-ttd (approved):

PR #278, opened 2024-05-30
UID2-1983 ensure first key in a keyset is immediately active

jon8787:


src/test/java/com/uid2/admin/vertx/EncryptionKeyServiceTest.java
```diff
@@ -177,21 +172,41 @@ void addSiteKeyUsesKeysetAndAddsKey() throws Exception {
...
         final KeysetKey key = keyService.addKeysetKey(1);
         verify(keysetKeyStoreWriter).upload(collectionOfSize(1), eq(124));
-        assertSiteKeyActivation(key, clock.now());
+
+        final int siteKeyActivatesInSeconds = 0; //since this is the first key for a keyset, test that it is immediately active
+        assertKeyActivation(clock.now(), siteKeyActivatesInSeconds, SITE_KEY_EXPIRES_AFTER_SECONDS, key.getCreated(), key.getActivates(), key.getExpires());
     }
 
+    @Test
+    void whenAddSecondKeyForKeysetItIsNotImmediatelyActive() throws Exception {
```
caroline-ttd, line 189: Nit pick, do we want to have one test case that when `isDuringRotation` is true?

caroline-ttd:
Over LGTM.

caroline-ttd (approved):

src/test/java/com/uid2/admin/vertx/EncryptionKeyServiceTest.java
jon8787: There are lots of existing rotation checks below. These call `checkRotatedKeyResponse` which ensures `activates` has the correct timestamp.

src/test/java/com/uid2/admin/vertx/EncryptionKeyServiceTest.java
```diff
@@ -177,21 +172,41 @@ void addSiteKeyUsesKeysetAndAddsKey() throws Exception {
...
+        }};
+        setAdminKeysets(keysets);
+
+        final KeysetKey[] keys = {
+                new KeysetKey(11, null, Instant.ofEpochMilli(KEY_CREATE_TIME_IN_MILLI), Instant.ofEpochMilli(KEY_ACTIVATE_TIME_IN_MILLI), Instant.ofEpochMilli(KEY_EXPIRE_TIME_IN_MILLI), keysetId),
+        };
+        setKeysetKeys(MAX_KEY_ID, keys);
+
+        final KeysetKey key2 = keyService.addKeysetKey(keysetId);
+        assertTrue(key2.getActivates().isAfter(key2.getCreated()));
```
sunnywu, line 202: should we check if activates is after created _but within_ certain expected time limit?

jon8787: I fixed the test to be more specific

sunnywu (approved):

PR #291, opened 2024-06-20
Participant Summary Tab

ssundahlTTD:
Add new on call tab for Participant Summary

### DIFF
--- a/webroot/adm/oncall/search.html
+++ b/webroot/adm/oncall/search.html
@@ -2,8 +2,7 @@
 
 <head>
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.1/jquery.min.js"></script>
-    <script src="/js/main.js"></script>
-    <link href="/css/bootstrap.min.css" rel="stylesheet">
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-QWTKZyjpPEjISv5WaRU9OFeRpok6YctnYmDr5pNlyT2bRjXh0JMhjY6hW+ALEwIH" crossorigin="anonymous">
     <script src="/js/bootstrap.bundle.min.js"></script>
 </head>
 
@@ -50,7 +49,7 @@
 
         </div>
     </div>
-    <script language="JavaScript">
+    <script>
         $(document).ready(function () {
             $('#doSearch').on('click', function () {
                 var keyString = $('#key').val();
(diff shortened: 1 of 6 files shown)


webroot/js/participantSummary.js
```diff
@@ -0,0 +1,129 @@
+let siteList;
+
+/* ***** multi-use error handler *** */
+function participantSummaryErrorHandler(err, divContainer) {
+    const errorMessage = prettifyJson(err.responseText);
+    $(divContainer).html(errorMessage).show();
+};
+
+function loadAllSitesCallback(result) {
+    siteList = JSON.parse(result).map(function(item) { return { name: item.name, id: item.id } });
```
alex-yau-ttd, line 10: nit: is there a good reason to use this notation over arrow notation? e.g.
```
siteList = JSON.parse(result).map((item) => { return { name: item.name, id: item.id } });
```
I think arrow notation is easier to read

webroot/adm/oncall/search.html
```diff
@@ -69,6 +68,8 @@ <h5>Search Results:</h5>
             }
         });
     </script>
+    <script src="../../js/main.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/js/bootstrap.bundle.min.js" integrity="sha384-YvpcrYf0tY3lHB60NNkmXc5s9fDVZLESaAA55NDzOxhy9GkcIdslK1eN7N6jIeHz" crossorigin="anonymous"></script>
```
alex-yau-ttd, line 72: My understanding is that putting the script tags here at the end of the body are better for performance. Should we consider doing this in the other files as well that use bootstrap as well? i.e. `generate-api-key-secret.html` and `generate-cstg-keypair.html`?

alex-yau-ttd (approved):

webroot/js/participantSummary.js
ssundahlTTD: Yeah, I originally used the old notation just for consistency.  But since this is a new file I think you're right.  I changed them all.

webroot/adm/oncall/search.html
ssundahlTTD: it looks like those scripts are already being loaded at the bottom of the body.

alex-yau-ttd: ah true. I think I got mixed up with the `main.js` script tag being in the `<head>` instead in a bunch of other files (e.g. [client-key.html](https://github.com/IABTechLab/uid2-admin/blob/main/webroot/adm/client-key.html#L4). Probably not worth changing though.

PR #308, opened 2024-06-26
Wzh UID2 3570 write encrypted s3files

lizk886:
Main Changes:
Image description
config, a flag to determine if s3 encryption is turned on
S3KeyManger, added to make sure every siteId has at least required number of S3 keys
PrivateSiteDataRefreshService, added EncryptedFilesSyncJob to its endpoint

Require shared branch https://github.com/IABTechLab/uid2-shared/pull/252


src/main/java/com/uid2/admin/store/writer/ScopedStoreWriter.java
```diff
@@ -59,7 +59,11 @@ public void upload(String data, JsonObject extraMeta) throws Exception {
         fileManager.uploadFile(location, dataFile, data);
         fileManager.uploadMetadata(metadata.getJson(), dataType, scope.getMetadataPath());
 
-        provider.loadContent(provider.getMetadata());
+        //The loadContent method is trying to read and parse this data as if it were unencrypted
+        // which is causing the deserialization error for encrypted json
+
+       // provider.loadContent(provider.getMetadata());
+
```
lizk886, line 66: But commenting this line out would fail a certain tests
Screenshot 2024-06-30 at 2 59 45 AM

src/test/java/com/uid2/admin/store/writer/EncryptedScopedStoreWriterTest.java
```diff
@@ -0,0 +1,385 @@
...
+    @Test
+    void testUploadLogic() throws Exception {
+        Collection<Site> sites = oneSite;
+        String data = jsonWriter.writeValueAsString(sites);
+        JsonObject extraMeta = new JsonObject();
+
+        // Manually execute the logic from upload method
+        if (isEncrypted(data)) {
+            // If it's encrypted, simulate calling the superclass upload method
+            System.out.println("Calling super.upload with already encrypted data");
```
thomasm-ttd, line 146: Is this a test failure? Printing to system out doesn't tell me if the tests passed or failed.

src/test/java/com/uid2/admin/store/writer/EncryptedScopedStoreWriterTest.java
```diff
@@ -0,0 +1,385 @@
...
+                if (key.getSiteId() == testSiteId) {
+                    if (largestKey == null || key.getId() > largestKey.getId()) {
+                        largestKey = key;
+                    }
+                }
+            }
+
+            if (largestKey != null) {
+                // Directly execute encryption logic here
+                byte[] secret = Base64.getDecoder().decode(largestKey.getSecret());
```
thomasm-ttd, line 166: Why are you duplicating the encryption logic here? Why doesn't the test call the uploadWithEncryptionKey method? We are trying to validate that the class works as expected, so we need to test the methods on the class

src/test/java/com/uid2/admin/store/writer/EncryptedScopedStoreWriterTest.java
```diff
@@ -0,0 +1,385 @@
...
+                clock,
+                scope,
+                dataFile,
+                dataType,
+                s3KeyProvider
+        );
+    }
+
+    @Test
+    void testEncryptData() throws Exception {
```
thomasm-ttd, line 122: This test doesn't actually call any methods on encryptedScopedStoreWriter - which is the target of the tests. 
Duplicating the logic and then testing that doesn't help us test that the class does what is expected.

src/test/java/com/uid2/admin/store/writer/EncryptedScopedStoreWriterTest.java
```diff
@@ -0,0 +1,385 @@
...
+        String encryptedFilePathStr = encryptedFilePath.toString();
+        if (files.contains(encryptedFilePathStr)) {
+            cloudStorage.delete(encryptedFilePathStr);
+        }
+
+        // Call fileManager to upload the encrypted data
+        fileManager.uploadFile(encryptedFilePath, dataFile, encryptedJson.encodePrettily());
+
+        // Verify the result
+        assertThat(cloudStorage.list(sitesDir)).contains(sitesDir + "/sites_encrypted.json");
```
thomasm-ttd, line 229: This is only one of the asserts I would expect - either in this test or another one. I would expect to see that the file that has been written contains the correct data, and that the data can be decrypted correctly.

thomasm-ttd (commented):
Have a look through the comments, and reach out if you want me to go through them with you on a call

src/main/java/com/uid2/admin/store/MultiScopeStoreWriter.java
```diff
@@ -62,4 +81,14 @@ public static <K, V> boolean areMapsEqual(Map<K, V> a, Map<K, V> b) {
     public static <T> boolean areCollectionsEqual(Collection<T> a, Collection<T> b) {
         return a.size() == b.size() && a.stream().allMatch(b::contains);
     }
-}
+
+    public void uploadGeneral(Map<Integer, T> desiredState, JsonObject extraMeta) throws Exception {
```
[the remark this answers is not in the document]
lizk886: That makes sense! Will change it

src/main/java/com/uid2/admin/Main.java
```diff
@@ -204,15 +205,16 @@ public void run() {
...
+                if (e.getMessage().contains("s3 get error") || e.getMessage().contains("The specified key does not exist")) {
                     s3KeyStoreWriter.upload(new HashMap<>(), null);
                     s3KeyProvider.loadContent();
                 } else {
                     throw e;
                 }
             }
             s3KeyManager.generateKeysForOperators(operatorKeyProvider.getAll(), config.getLong("s3_key_activates_in_seconds"), config.getInteger("s3_key_count_per_site"));
-
+            Thread.sleep(8000); // Wait before checking again
```
lizk886, line 216: Issue: When admin get started for the first time, some times it takes a few times of running to fully generate the whole s3 encryption keys file

src/main/java/com/uid2/admin/store/MultiScopeStoreWriter.java
```diff
@@ -42,16 +50,27 @@ private Map<Integer, T> getCurrentState(Collection<Integer> siteIds) throws Exce
         for (Integer siteId : siteIds) {
             StoreReader<T> reader = factory.getReader(siteId);
             if (fileManager.isPresent(reader.getMetadataPath())) {
-                reader.loadContent();
-                currentState.put(siteId, reader.getAll());
+                try {
+                    reader.loadContent();
+                    currentState.put(siteId, reader.getAll());
+                } catch (Exception e) {
```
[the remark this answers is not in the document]
lizk886: Those are all absolute valuable reminders! Since now I isolated plain text and encrypted logics, I no longer need this try catch block

src/main/java/com/uid2/admin/store/writer/ScopedStoreWriter.java
lizk886, line 1: need a test for this code change

src/main/java/com/uid2/admin/Main.java
```diff
@@ -204,15 +205,13 @@ public void run() {
             try {
                 s3KeyProvider.loadContent();
             } catch (CloudStorageException e) {
-                if (e.getMessage().contains("s3 get error")) {
+                if (e.getMessage().contains("s3 get error") || e.getMessage().contains("The specified key does not exist")) {
```
Ian-Nara, line 208: does "s3 get error" definitely mean the file doesn't exist in s3? Worried that if we can't reach s3 for some other reason, this will overwrite the existing file.

src/main/java/com/uid2/admin/job/jobsync/acl/KeyAclSyncJob.java
```diff
@@ -32,9 +34,12 @@ public String getId() {
 
     @Override
     public void execute() throws Exception {
-        HashMap<Integer, Map<Integer, EncryptionKeyAcl>> desiredState = PrivateSiteUtil.getEncryptionKeyAclsForEachSite(
-                globalOperators,
-                globalAcls);
-        multiScopeStoreWriter.uploadIfChanged(desiredState, null);
+        HashMap<Integer, Map<Integer, EncryptionKeyAcl>> desiredPrivateState = PrivateSiteUtil.getEncryptionKeyAclsForEachSite(globalOperators, globalAcls);
```
cody-constine-ttd, line 37: These jobs are all called by private site data. I think we need to make new versions of them, so we don't interfere with it.

src/main/java/com/uid2/admin/store/MultiScopeStoreWriter.java
```diff
@@ -18,12 +25,29 @@ public MultiScopeStoreWriter(FileManager fileManager, StoreFactory<T> factory, B
         this.areEqual = areEqual;
     }
 
+    public void uploadWithEncryptionOrChanges(Map<Integer, T> desiredState, JsonObject extraMeta) throws Exception {
```
cody-constine-ttd, line 28: When we separate the jobs, lets just change this to just encryption.

src/main/java/com/uid2/admin/store/writer/EncryptedScopedStoreWriter.java
```diff
@@ -0,0 +1,113 @@
...
+                .put("key_id", encryptionKey.getId())
+                .put("encryption_version", "1.0")
+                .put("encrypted_payload", Base64.getEncoder().encodeToString(encryptedPayload));
+
+        this.upload(encryptedJson.encodePrettily(), extraMeta);
+    }
+
+    @Override
+    public void upload(String data, JsonObject extraMeta) throws Exception {
+        if (isEncrypted(data)) {
```
cody-constine-ttd, line 50: What is the case this will get called with already encrypted data? Your logic is a bit convoluted here. It would be clearer if we assume this is called with unencrypted data and then encrypt it. Then just make the above function private.

src/main/java/com/uid2/admin/store/writer/EncryptedScopedStoreWriter.java
```diff
@@ -0,0 +1,113 @@
...
+            return json.containsKey("key_id") &&
+                    json.containsKey("encryption_version") &&
+                    json.containsKey("encrypted_payload");
+        } catch (Exception e) {
+            return false;
+        }
+    }
+
+
+    public String getDecryptedContent(String encryptedContent) throws Exception {
```
cody-constine-ttd, line 88: Why does the writer need the ability to decrypt?

src/main/java/com/uid2/admin/store/writer/ScopedStoreWriter.java
```diff
@@ -59,7 +64,20 @@ public void upload(String data, JsonObject extraMeta) throws Exception {
         fileManager.uploadFile(location, dataFile, data);
         fileManager.uploadMetadata(metadata.getJson(), dataType, scope.getMetadataPath());
 
-        provider.loadContent(provider.getMetadata());
+        if (!isEncrypted(data)) {
```
cody-constine-ttd, line 67: Not sure I really understand what is happening here either.

src/main/java/com/uid2/admin/job/jobsync/acl/KeyAclSyncJob.java
lizk886: sounds good!

src/main/java/com/uid2/admin/store/writer/EncryptedScopedStoreWriter.java
lizk886: I have this decryption function here for testing

src/main/java/com/uid2/admin/store/writer/ScopedStoreWriter.java
lizk886: After uploading data, provider.loadContents() is called to refresh the file contents. However, this cannot be applied to encrypted data because they cannot be deserialized into any data type.This check helps determine if we are uploading a plaintext file or an encrypted file. If it's an encrypted file, we skip the refresh to avoid these issues.

src/main/java/com/uid2/admin/Main.java
lizk886: I don't think I know the exact answer! Probably need more input from the team

lizk886: Will make a new pr in shared to solve this

src/main/java/com/uid2/admin/store/writer/ScopedStoreWriter.java
lizk886: will make a new encrypted scope store reader to decrypt before load contents

src/main/java/com/uid2/admin/store/writer/EncryptedScopedStoreWriter.java
lizk886: will move decryption logics to shared

conf/local-e2e-docker-config.json
```diff
@@ -34,6 +34,7 @@
   "salt_snapshot_expires_after_seconds": 180000,
   "is_auth_disabled": true,
   "enable_keysets": true,
+  "enable_s3_encryption": true,
```
thomasm-ttd, line 37: Best to keep all the s3 encryption config values together, so please move this to next to the other settings in all the config files

src/main/java/com/uid2/admin/legacy/LegacyClientKeyStoreWriter.java
```diff
@@ -24,6 +26,18 @@ public LegacyClientKeyStoreWriter(RotatingLegacyClientKeyProvider provider, File
...
+                                      FileManager fileManager,
+                                      ObjectWriter jsonWriter,
+                                      VersionGenerator versionGenerator,
+                                      Clock clock,
+                                      StoreScope scope,
+                                      RotatingS3KeyProvider s3KeyProvider) {
+        this.jsonWriter = jsonWriter;
+        FileName dataFile = new FileName("clients", ".json");
+        String dataType = "client_keys";
+        this.writer = new EncryptedScopedStoreWriter(provider, fileManager, versionGenerator, clock, scope, dataFile, dataType, s3KeyProvider, scope.getId());
```
thomasm-ttd, line 39: Are you 100% sure this will never be a globalscope? That returns a null for getId().

src/main/java/com/uid2/admin/store/MultiScopeStoreWriter.java
```diff
@@ -49,17 +61,33 @@ private Map<Integer, T> getCurrentState(Collection<Integer> siteIds) throws Exce
...
     }
 
-    private void write(Map<Integer, T> desiredState, Collection<Integer> sitesToWrite, JsonObject extraMeta) throws Exception {
-        for (Integer addedSite : sitesToWrite) {
-            factory.getWriter(addedSite).upload(desiredState.get(addedSite), extraMeta);
+    public void uploadEncrypted(Map<Integer, T> desiredState, JsonObject extraMeta) throws Exception {
+        EncryptedStoreFactory<T> encryptedFactory = (EncryptedStoreFactory<T>) factory;
+        for (Map.Entry<Integer, T> entry : desiredState.entrySet()) {
+            Integer siteId = entry.getKey();
+            if (siteId != null) {
```
thomasm-ttd, line 68: When would siteId be null? What should be done in this case?

src/test/java/com/uid2/admin/util/PublicSiteUtilTest.java
```diff
@@ -0,0 +1,319 @@
...
+                            .build(),
+                    new OperatorBuilder()
+                            .withSiteId(siteId2)
+                            .withType(OperatorType.PUBLIC)
+                            .build()
+            );
+
+            Set<Site> sites = ImmutableSet.of(
+                    new Site(siteId1, "Site 1", true),
+                    new Site(siteId2, "Site 2", true)
```
cody-constine-ttd, line 90: Lets add a third site to this test to be sure

src/main/java/com/uid2/admin/legacy/LegacyClientKeyStoreWriter.java
lizk886: Global scope would only be passed into this legacyClientKeyStoreWriter constructor during main and globalWriter in dataTypeFactories. I have checked and don't think globalScope would be passed in with this constructor having S3 key provider by any chance. This constructor with s3 key provider is only called in dataTypeFactories getEncryptedWriter function, which passed in encryptedScope only. I will change this constructor from taking ScopedStore to EncryptedScopeStorage, in case there are any mismatch that I missed.

lizk886: Now all the constructors that will initiate EncryptedScopedStoreWriter takes in EncryptedScope instead of ScopedStore. As long as the program runs without errors, that should verify no globalScope indeed come at this constuctor.

src/main/java/com/uid2/admin/store/MultiScopeStoreWriter.java
lizk886: This siteId would never be null, as ensured by getPublicSitesMap in PublicSiteUtil. I can remove this check if thats better

src/test/java/com/uid2/admin/job/sitesync/SiteEncryptionTest.java
lizk886, line 1: still working on this one

src/main/java/com/uid2/admin/store/MultiScopeStoreWriter.java
```diff
@@ -49,9 +61,19 @@ private Map<Integer, T> getCurrentState(Collection<Integer> siteIds) throws Exce
         return currentState;
     }
 
-    private void write(Map<Integer, T> desiredState, Collection<Integer> sitesToWrite, JsonObject extraMeta) throws Exception {
-        for (Integer addedSite : sitesToWrite) {
-            factory.getWriter(addedSite).upload(desiredState.get(addedSite), extraMeta);
+    public void uploadEncrypted(Map<Integer, T> desiredState, JsonObject extraMeta) throws Exception {
```
thomasm-ttd, line 64: Can you rename this to uploadPrivateWithEncryption to match the uploadPublicWithEncryption

thomasm-ttd (commented):
Some small changes, but overall looking good

cody-constine-ttd (approved):

PR #311, opened 2024-07-03
Check duplicate enclave

thomasm-ttd:
Added validations for the admin Enclave Id endpoints.
Check for duplicate enclave Id and protocol.
Added unit tests for all the end points

### DIFF
--- a/src/main/java/com/uid2/admin/vertx/service/EnclaveIdService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/EnclaveIdService.java
@@ -2,6 +2,7 @@ package com.uid2.admin.vertx.service;
 
 import com.fasterxml.jackson.databind.ObjectWriter;
 import com.uid2.admin.auth.AdminAuthMiddleware;
+import com.uid2.admin.store.Clock;
 import com.uid2.admin.store.writer.EnclaveStoreWriter;
 import com.uid2.admin.vertx.JsonUtil;
 import com.uid2.admin.vertx.RequestUtil;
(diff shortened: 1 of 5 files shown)


src/main/java/com/uid2/admin/vertx/service/EnclaveIdService.java
```diff
@@ -95,35 +95,56 @@ private void handleEnclaveAdd(RoutingContext rc) {
...
+            final String enclaveId = rc.queryParam("enclave_id").stream().findFirst().orElse(null);
+            if (enclaveId == null || enclaveId == "") {
                 ResponseUtil.error(rc, 400, "enclave_id not specified");
                 return;
             }
 
-            List<EnclaveIdentifier> enclaveIds = this.enclaveIdProvider.getAll()
-                    .stream().sorted((a, b) -> (int) (a.getCreated() - b.getCreated()))
-                    .collect(Collectors.toList());
+            long created = Instant.now().getEpochSecond();
```
cYKatherine, line 122: nit: Can we use https://github.com/IABTechLab/uid2-admin/blob/main/src/main/java/com/uid2/admin/store/Clock.java instead of `Instant.now()` just to be consistent across the repo. We use the customised Clock so that we are able to mock those values in the tests.

cYKatherine (approved):

src/main/java/com/uid2/admin/vertx/service/EnclaveIdService.java
thomasm-ttd: Done

PR #329, opened 2024-08-05
UID2-3506 Add functionality on oncall page to list related keysets and rotate them

cYKatherine:
UI looks below. It will highlight the reason why the keyset is related to the participant. Also tested the rotation works.

image

image

image

### DIFF
--- a/src/main/java/com/uid2/admin/Main.java
+++ b/src/main/java/com/uid2/admin/Main.java
@@ -251,7 +251,7 @@ public class Main {
                     new EnclaveIdService(auth, writeLock, enclaveStoreWriter, enclaveIdProvider, clock),
                     encryptionKeyService,
                     new KeyAclService(auth, writeLock, keyAclStoreWriter, keyAclProvider, siteProvider, encryptionKeyService),
-                    new SharingService(auth, writeLock, adminKeysetProvider, keysetManager, siteProvider, enableKeysets),
+                    new SharingService(auth, writeLock, adminKeysetProvider, keysetManager, siteProvider, enableKeysets, clientKeyProvider),
                     clientSideKeypairService,
                     new ServiceService(auth, writeLock, serviceStoreWriter, serviceProvider, siteProvider, serviceLinkProvider),
                     new ServiceLinkService(auth, writeLock, serviceLinkStoreWriter, serviceLinkProvider, serviceProvider, siteProvider),
--- a/src/main/java/com/uid2/admin/vertx/service/SharingService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/SharingService.java
@@ -183,10 +183,10 @@ public class SharingService implements IService {
                 return;
             }
 
-            // Get value for client type
+            // Get value for client type from the backend
             Set<ClientType> clientTypes = this.siteProvider.getSite(siteId).getClientTypes();
 
-            // Check if this site has any cleint key that has an ID_READER role
+            // Check if this site has any client key that has an ID_READER role
             boolean isIdReaderRole = false;
             for (LegacyClientKey c : this.clientKeyProvider.getAll()) {
                 if (c.getRoles().contains(Role.ID_READER)) {
@@ -199,10 +199,10 @@ public class SharingService implements IService {
             Map<Integer, AdminKeyset> collection = this.keysetProvider.getSnapshot().getAllKeysets();
             for (Map.Entry<Integer, AdminKeyset> keyset : collection.entrySet()) {
                 // The keysets meet any of the below conditions ALL need to be rotated:
-                // a. Keysets where allowed_types include any of the clientTypes that you noted down earlier
-                // b. If leaked key was an ID_READER, we want to rotate the keysets where allowed_sites is set to null
+                // a. Keysets where allowed_types include any of the clientTypes of the site
+                // b. If this participant has a client key with ID_READER role, we want to rotate all the keysets where allowed_sites is set to null
                 // c. Keysets where allowed_sites include the leaked site
-                // d. Keysets belonging to the leaked site itself (can also get these keysets by putting down the "Site id" and click "List All Keysets By Site Id")
+                // d. Keysets belonging to the leaked site itself
                 if (containsAny(keyset.getValue().getAllowedTypes(), clientTypes) ||
                         isIdReaderRole && keyset.getValue().getAllowedSites() == null ||
                         keyset.getValue().getAllowedSites() != null && keyset.getValue().getAllowedSites().contains(siteId) ||
--- a/src/main/java/com/uid2/admin/vertx/service/SharingService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/SharingService.java
@@ -179,10 +179,7 @@ public class SharingService implements IService {
             }
 
             // Get value for client type
-            Set<ClientType> clientTypes = new HashSet<>();
-            if (!rc.queryParam("client_types").isEmpty()) {
-                clientTypes = getTypes(rc.queryParam("client_types").get(0));
-            }
+            Set<ClientType> clientTypes = this.siteProvider.getSite(siteId).getClientTypes();
 
 //            // Check if the key has a ID_READER role
             boolean isIdReaderRole = true;
--- a/src/main/java/com/uid2/admin/vertx/service/SharingService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/SharingService.java
@@ -165,10 +165,13 @@ public class SharingService implements IService {
         try {
             // Get value for site id
             final Optional<Integer> siteIdOpt = RequestUtil.getSiteId(rc, "site_id");
-            if (!siteIdOpt.isPresent()) return;
+            if (!siteIdOpt.isPresent()) {
+                ResponseUtil.error(rc, 400, "must specify a site id");
+                return;
+            }
             final int siteId = siteIdOpt.get();
 
-            if (siteId != Const.Data.AdvertisingTokenSiteId && !SiteUtil.isValidSiteId(siteId)) {
+            if (!SiteUtil.isValidSiteId(siteId)) {
                 ResponseUtil.error(rc, 400, "must specify a valid site id");
                 return;
             }
--- a/src/main/java/com/uid2/admin/vertx/service/SharingService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/SharingService.java
@@ -189,7 +189,7 @@ public class SharingService implements IService {
             // Check if this site has any cleint key that has an ID_READER role
             boolean isIdReaderRole = false;
             for (LegacyClientKey c : this.clientKeyProvider.getAll()) {
-                if (c.getRoles().contains("ID_READER")) {
+                if (c.getRoles().contains(Role.ID_READER)) {
                     isIdReaderRole = true;
                 }
             }
--- a/src/main/java/com/uid2/admin/vertx/service/SharingService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/SharingService.java
@@ -2,6 +2,7 @@ package com.uid2.admin.vertx.service;
 
 import com.uid2.admin.auth.AdminAuthMiddleware;
 import com.uid2.admin.auth.AdminKeyset;
+import com.uid2.admin.legacy.LegacyClientKey;
 import com.uid2.admin.store.reader.RotatingAdminKeysetStore;
 import com.uid2.admin.vertx.RequestUtil;
 import com.uid2.admin.vertx.WriteLock;
@@ -10,6 +11,7 @@ import com.uid2.admin.vertx.ResponseUtil;
 import com.uid2.shared.Const;
 import com.uid2.shared.auth.Role;
 import com.uid2.shared.model.ClientType;
+import com.uid2.shared.model.Site;
 import com.uid2.shared.model.SiteUtil;
 import com.uid2.shared.store.reader.RotatingSiteStore;
 import io.vertx.core.http.HttpHeaders;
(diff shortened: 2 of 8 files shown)


webroot/js/participantSummary.js
```diff
@@ -123,7 +148,21 @@ $(document).ready(() => {
...
         $('.section').show();
     });
+
+    $('#doRotateKeysets').on('click', () => {
+        var keysets = $('#relatedKeysetsStandardOutput').text();
+        console.log(keysets);
+        const ja = JSON.parse(keysets);
+        ja.forEach((keyset) => {
+            var url = `/api/key/rotate_keyset_key?min_age_seconds=3600&keyset_id=${keyset.keyset_id}&force=true`;
+//            if ($('#force').is(':checked')) url = url + '&force=true';
```
cYKatherine, line 163: I think the rotation only works when we set `force` to true. Does anyone have any extra insight to this?

src/main/resources/localstack/s3/core/keysets/keysets.json
```diff
@@ -89,5 +89,23 @@
     "created" : 1617149276,
     "enabled" : true,
     "default" : true
+  },
+  {
+    "keyset_id": 1001,
```
cYKatherine, line 94: Leave it here for testing purpose, will revert before merging

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -150,6 +156,66 @@ private void handleSetKeyset(RoutingContext rc) {
...
+            }
+
+            // Get value for client type
+            Set<ClientType> clientTypes = new HashSet<>();
+            if (!rc.queryParam("client_types").isEmpty()) {
+                clientTypes = getTypes(rc.queryParam("client_types").get(0));
+            }
+
+//            // Check if the key has a ID_READER role
+            boolean isIdReaderRole = true;
```
cYKatherine, line 188: I set this to constant `true`. Question here is: if we need to rotate related keysets for a participant, do we care whether the leaked key is ID_READER or not? 

My take is we will just rotate regardless of type of the leaked key. WDYT?

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -150,6 +156,66 @@ private void handleSetKeyset(RoutingContext rc) {
...
+
+            if (siteId != Const.Data.AdvertisingTokenSiteId && !SiteUtil.isValidSiteId(siteId)) {
+                ResponseUtil.error(rc, 400, "must specify a valid site id");
+                return;
+            }
+
+            // Get value for client type
+            Set<ClientType> clientTypes = new HashSet<>();
+            if (!rc.queryParam("client_types").isEmpty()) {
+                clientTypes = getTypes(rc.queryParam("client_types").get(0));
```
cYKatherine, line 184: Get this from the backend

src/main/java/com/uid2/admin/Main.java
```diff
@@ -251,7 +251,7 @@ public void run() {
                     new EnclaveIdService(auth, writeLock, enclaveStoreWriter, enclaveIdProvider, clock),
                     encryptionKeyService,
                     new KeyAclService(auth, writeLock, keyAclStoreWriter, keyAclProvider, siteProvider, encryptionKeyService),
-                    new SharingService(auth, writeLock, adminKeysetProvider, keysetManager, siteProvider, enableKeysets),
+                    new SharingService(auth, writeLock, adminKeysetProvider, keysetManager, siteProvider, enableKeysets, clientKeyProvider),
```
cYKatherine, line 254: Need this because we need to know whether the site/participant has any client key that have the `ID_READER` role

src/test/java/com/uid2/admin/vertx/ClientKeyServiceTest.java
```diff
@@ -460,7 +460,7 @@ private static void assertAddedClientKeyEquals(ClientKey expected, ClientKey act
                 .isEqualTo(expected);
     }
 
-    private static class LegacyClientBuilder {
+    public static class LegacyClientBuilder {
```
cYKatherine, line 463: Change this to public as needed in the `SharingServiceTest`

webroot/js/participantSummary.js
```diff
@@ -7,7 +7,7 @@ function participantSummaryErrorHandler(err, divContainer) {
 };
 
 function loadAllSitesCallback(result) {
-    siteList = JSON.parse(result).map((item) => { return { name: item.name, id: item.id } });
+    siteList = JSON.parse(result).map((item) => { return { name: item.name, id: item.id, clientTypes: item.clientTypes } });
```
cYKatherine, line 10: Still need this from the front end as I want to highlight the clientTypes in the UI.

webroot/js/participantSummary.js
```diff
@@ -88,6 +88,40 @@ function loadOptoutWebhooksCallback(result, siteName) {
     $('#webhooksStandardOutput').html(formatted);
 };
 
+function loadRelatedKeysetsCallback(result, siteId, clientTypes) {
+    const resultJson = JSON.parse(result);
```
cYKatherine, line 92: See how it looks like in the description screenshot

webroot/js/participantSummary.js
```diff
@@ -123,7 +164,23 @@ $(document).ready(() => {
...
         $('.section').show();
     });
+
+    $('#doRotateKeysets').on('click', () => {
+        var keysets = $('#relatedKeysetsStandardOutput').text();
+        const ja = JSON.parse(keysets);
+        var rotateKeysetsMessage = '';
+        ja.forEach((keyset) => {
+            var url = `/api/key/rotate_keyset_key?min_age_seconds=3600&keyset_id=${keyset.keyset_id}&force=true`;
+            doApiCallWithCallback(
```
cYKatherine, line 179: See result of this in the PR description screenshot. 

Each time a keyset is rotated, the rotated keyset information will show on the UI. cc: @vishalegbert-ttd

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -150,6 +161,64 @@ private void handleSetKeyset(RoutingContext rc) {
...
+    }
+
+    private void handleListAllKeysetsRelated(RoutingContext rc) {
+        try {
+            // Get value for site id
+            final Optional<Integer> siteIdOpt = RequestUtil.getSiteId(rc, "site_id");
+            if (!siteIdOpt.isPresent()) return;
+            final int siteId = siteIdOpt.get();
+
+            if (siteId != Const.Data.AdvertisingTokenSiteId && !SiteUtil.isValidSiteId(siteId)) {
```
vishalegbert-ttd, line 181: We can check if the site ID is present in the request as well? Right now we just return?

webroot/js/participantSummary.js
```diff
@@ -123,7 +164,23 @@ $(document).ready(() => {
...
+        doApiCallWithCallback('GET', url, (r) => { loadRelatedKeysetsCallback(r, site.id, site.clientTypes) }, (err) => { participantSummaryErrorHandler(err, '#relatedKeysetsErrorOutput') });
         $('.section').show();
     });
+
+    $('#doRotateKeysets').on('click', () => {
+        var keysets = $('#relatedKeysetsStandardOutput').text();
+        const ja = JSON.parse(keysets);
+        var rotateKeysetsMessage = '';
+        ja.forEach((keyset) => {
+            var url = `/api/key/rotate_keyset_key?min_age_seconds=3600&keyset_id=${keyset.keyset_id}&force=true`;
```
vishalegbert-ttd, line 178: I doublechecked the code, the min age seconds doesn't matter if we set force to true, maybe we can just set min_age_seconds=0 here

vishalegbert-ttd:
Just want to confirm, it looks like the Rotate All Keys will also rotate master keys and site keys for ID_READER?

cYKatherine:
> Just want to confirm, it looks like the Rotate All Keys will also rotate master keys and site keys for ID_READER?

Technically speaking this shouldn't happen. master keys and site keys (with site id `-1` and `-2`) shouldn't have any `allowed_sites` or `allowed_types`. I checked our integ and prod environment and attached the screenshot.
image

webroot/js/participantSummary.js
cYKatherine: Yes, I tried setting `min_age_seconds=0` but it will return with 404 as it needs a positive integer 😢 I can change it to 1 if that makes more sense 🤷‍♀️

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -150,6 +161,64 @@ private void handleSetKeyset(RoutingContext rc) {
         }
     }
 
+    // Method to check if one set contains any values from another set
+    private static <T> boolean containsAny(Set<T> set1, Set<T> set2) {
+        for (T element : set2) {
```
jon8787, line 166: is this method necessary? https://stackoverflow.com/questions/8708542/something-like-contains-any-for-java-set

src/main/java/com/uid2/admin/vertx/service/SharingService.java
jon8787, line 181: `isValidSiteId` already checks that it's not AdvertisingTokenSiteId ?

src/main/java/com/uid2/admin/vertx/service/SharingService.java
```diff
@@ -150,6 +161,64 @@ private void handleSetKeyset(RoutingContext rc) {
...
+            // Get the keyset ids that need to be rotated
+            final JsonArray ja = new JsonArray();
+            Map<Integer, AdminKeyset> collection = this.keysetProvider.getSnapshot().getAllKeysets();
+            for (Map.Entry<Integer, AdminKeyset> keyset : collection.entrySet()) {
+                // The keysets meet any of the below conditions ALL need to be rotated:
+                // a. Keysets where allowed_types include any of the clientTypes of the site
+                // b. If this participant has a client key with ID_READER role, we want to rotate all the keysets where allowed_sites is set to null
+                // c. Keysets where allowed_sites include the leaked site
+                // d. Keysets belonging to the leaked site itself
+                if (containsAny(keyset.getValue().getAllowedTypes(), clientTypes) ||
```
jon8787, line 206: Btw there is a method [canClientAccessKey](https://github.com/IABTechLab/uid2-shared/blob/19edb010c6a4d753d03c89268c238be10a8f6722/src/main/java/com/uid2/shared/auth/KeysetSnapshot.java#L13) which is the source of truth. 

Would be great if we could use that here, although might not be feasible. If we can't use it, still worth mentioning it here as a comment.

src/test/java/com/uid2/admin/vertx/SharingServiceTest.java
```diff
@@ -1259,4 +1255,153 @@ void KeysetSetNewWithType(Vertx vertx, VertxTestContext testContext) {
             testContext.completeNow();
         });
     }
+
+    @Test
+    void RelatedKeysetSetsWithClientTypes(Vertx vertx, VertxTestContext testContext) {
+        fakeAuth(Role.MAINTAINER);
+
+        AdminKeyset adminKeyset1 = new AdminKeyset(3, 5, "test", Set.of(4,6,7), Instant.now().getEpochSecond(),true, true, new HashSet<>(Arrays.asList(ClientType.DSP, ClientType.PUBLISHER)));
```
jon8787, line 1263: might be worth checking out the test cases for [canClientAccessKey](https://github.com/IABTechLab/uid2-shared/blob/19edb010c6a4d753d03c89268c238be10a8f6722/src/main/java/com/uid2/shared/auth/KeysetSnapshot.java#L13)

webroot/js/participantSummary.js
jon8787: should we make `min_age_seconds=0` not return a 404 then and get rid of force?

cYKatherine: Lemme separate that to a new ticket as it's behaviour change 😢 and need some extra frontend work

cYKatherine: https://atlassian.thetradedesk.com/jira/browse/UID2-3898

jon8787:
did we add some text / warning about what will happen when you click Rotate?

cYKatherine:
> did we add some text / warning about what will happen when you click Rotate?

Done

webroot/adm/oncall/participant-summary.html
```diff
@@ -82,6 +82,27 @@ <h5>Participant Opt-out Webhook</h5>
...
+                            a. Keysets where allowed_types include any of the clientTypes of the site<br>
+                            b. If this participant has a client key with ID_READER role, we want to rotate all the keysets where allowed_sites is set to null<br>
+                            c. Keysets where allowed_sites include the leaked site<br>
+                            d. Keysets belonging to the leaked site itself<br></div>
+                        <pre class="errorDiv" id="relatedKeysetsErrorOutput"></pre>
+                        <pre id="relatedKeysetsStandardOutput"></pre>
+
+                        <div class="col">
+                            <a href="#" class="btn btn-primary" id="doRotateKeysets">Rotate Keysets</a>
+                            <div>Note: Once rotated, the participant will need to get a new refresh key to be able to decrypt data.</div>
```
jon8787, line 98: Suggested wording: 
"Normally, keys don't become active for 24 hours when rotated, which gives participants 24 hours before they need to call sdk.refresh().
However in this case, rotation will make the new keys active immediately. This means participants will not be able to decrypt newly created UID tokens until they have called sdk.refresh(). Note that sdk.refresh() has a recommeded frequency of once per hour".

cYKatherine: image

PR #337, opened 2024-09-12
auto-complete for site in participant summary

ssundahlTTD:
https://github.com/user-attachments/assets/9bc5a889-f7a5-4c74-abee-c7e9b5a53299

### DIFF
--- a/webroot/adm/oncall/participant-summary.html
+++ b/webroot/adm/oncall/participant-summary.html
@@ -31,8 +31,7 @@
                                 <input type="text" class="form-control" id="site-search">
                                 <pre class="errorDiv" id="siteSearchErrorOutput"></pre>
                             </div>
-                            <div class="search-autocomplete-results dropdown-menu">
-                            </div>
+                            <div class="search-autocomplete-results dropdown-menu"></div>
                         </div>
                         <div class="col">
                             <a href="#" class="btn btn-primary search-button" id="doSearch">Search</a>
--- a/webroot/adm/oncall/participant-summary.html
+++ b/webroot/adm/oncall/participant-summary.html
@@ -4,6 +4,7 @@
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.1/jquery.min.js"></script>
     <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-QWTKZyjpPEjISv5WaRU9OFeRpok6YctnYmDr5pNlyT2bRjXh0JMhjY6hW+ALEwIH" crossorigin="anonymous">
     <link href="../../css/participantSummary.css" rel="stylesheet" />
+    <link rel="shortcut icon" href="#">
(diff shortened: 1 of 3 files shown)


webroot/adm/oncall/participant-summary.html
```diff
@@ -4,6 +4,7 @@
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.1/jquery.min.js"></script>
     <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-QWTKZyjpPEjISv5WaRU9OFeRpok6YctnYmDr5pNlyT2bRjXh0JMhjY6hW+ALEwIH" crossorigin="anonymous">
     <link href="../../css/participantSummary.css" rel="stylesheet" />
+    <link rel="shortcut icon" href="#">
```
alex-yau-ttd, line 7: Do we need this line?

webroot/js/participantSummary.js
```diff
@@ -10,6 +20,37 @@ function loadAllSitesCallback(result) {
...
+    document.querySelector('#site-search').value = searchText;
+    $(resultsElement).hide();
+    $('#doSearch').click();
+}
+
+function searchSitesAutocomplete(e) {
+    
+    searchString = searchInput.value;
+    const options = {
+        threshold: .2,
```
alex-yau-ttd, line 33: Curious - I see the default [threshold ](https://www.fusejs.io/api/options.html#threshold)is 0.6. Is there a good reason not to use that?

webroot/js/participantSummary.js
```diff
@@ -10,6 +20,37 @@ function loadAllSitesCallback(result) {
     siteList = JSON.parse(result).map((item) => { return { name: item.name, id: item.id, clientTypes: item.clientTypes } });
 };
 
+function setSearchValue(searchText) {
+    document.querySelector('#site-search').value = searchText;
+    $(resultsElement).hide();
+    $('#doSearch').click();
+}
+
+function searchSitesAutocomplete(e) {
```
alex-yau-ttd, line 29: do we need the `e` parameter? it's unused

alex-yau-ttd (approved):

webroot/adm/oncall/participant-summary.html
ssundahlTTD: This fixes a harmless but annoying 404 error in the web debug console.

webroot/js/participantSummary.js
ssundahlTTD: I intentionally set it lower because I didn't see the point in getting auto-complete suggestions that are really not close to what we're looking for.  like being multiple letters off.  but we can change this if we think it's too restrictive.

webroot/js/participantSummary.js
ssundahlTTD: we don't need it but I figured it's an okay reminder that the button that was pressed is available for use in case we want to do something like catch the enter key at some point.

PR #352, opened 2024-11-25
Adding a config file that makes workspace start better

cody-constine-ttd:


Ian-Nara:
Indentation format like this: https://github.com/IABTechLab/uid2-admin/blob/main/src/main/resources/localstack/s3/core/keys/keys.json?

Ian-Nara (approved):

PR #376, opened 2025-03-04
Store previous snapshots in encrypted

abuabraham-ttd:


pom.xml
```diff
@@ -10,13 +10,13 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <vertx.version>4.5.11</vertx.version>
+        <vertx.version>4.5.13</vertx.version>
         <vertx-maven-plugin.version>1.0.22</vertx-maven-plugin.version>
```
abuabraham-ttd, line 14: Upgrading versions similar to other applications for vulnerability fix

src/main/java/com/uid2/admin/store/writer/EncryptedSaltStoreWriter.java
```diff
@@ -91,6 +95,16 @@ protected void refreshProvider() {
...
+    @Override
+    protected List<RotatingSaltProvider.SaltSnapshot> getSnapshots(RotatingSaltProvider.SaltSnapshot data){
+    /*
+    Since metadata.json is overwritten during the process, we maintain a history of all snapshots seen so far.
+    On the final write, we append this history to metadata.json to ensure no snapshots are lost.
+    */
+        this.previousSeenSnapshots.add(data);
+        return this.previousSeenSnapshots;
+    }
+
```
abuabraham-ttd, line 107: Since in EncryptedSaltStoreWriter, we process salts/metadata.json and calls upload for each file separately, it updates encrypted/site_id/metadata.json with the file it is processing. Meaning, metadata.json will only contain 1 value, and that will be the last processed item. 

previousSeenSnapshots ensures, we always store previous seen and uses it. 

Admit; these are all retrofitting to use same upload method.

src/main/java/com/uid2/admin/store/writer/SaltStoreWriter.java
```diff
@@ -43,6 +43,20 @@ public SaltStoreWriter(JsonObject config, RotatingSaltProvider provider, FileMan
...
+                .collect(Collectors.toList());
+        RotatingSaltProvider.SaltSnapshot newestEffectiveSnapshot = currentSnapshots.stream()
+                .filter(snapshot -> snapshot.isEffective(now))
+                .reduce((a, b) -> b).orElse(null);
+        return Stream.concat(provider.getSnapshots().stream(), Stream.of(data))
+                .filter(snapshot -> newestEffectiveSnapshot == null || snapshot == newestEffectiveSnapshot)
+                .sorted(Comparator.comparing(RotatingSaltProvider.SaltSnapshot::getEffective))
+                .collect(Collectors.toList());
+     }
+
```
abuabraham-ttd, line 59: moved the condition out to separate method to override it

src/main/java/com/uid2/admin/store/writer/SaltStoreWriter.java
```diff
@@ -84,13 +90,6 @@ public void upload(RotatingSaltProvider.SaltSnapshot data) throws Exception {
                 continue;
             }
 
-            if (newestEffectiveSnapshot != null && snapshot != newestEffectiveSnapshot) {
```
cody-constine-ttd, line 87: Why did you remove this if? We now lost this info message

src/main/java/com/uid2/admin/store/writer/SaltStoreWriter.java
cody-constine-ttd: This logic doesn't look 1:1 with what you removed below. Looks like we are calculating newest effective twice. But not using it below anymore.

src/main/java/com/uid2/admin/store/writer/SaltStoreWriter.java
abuabraham-ttd: https://stackoverflow.com/questions/29571846/using-list-stream-filter-vs-for-loop

src/main/java/com/uid2/admin/store/writer/SaltStoreWriter.java
abuabraham-ttd: how? We are calculating in line 51, and filtering in 54

src/main/java/com/uid2/admin/store/writer/SaltStoreWriter.java
cody-constine-ttd: Right but we still lost the info message....

src/main/java/com/uid2/admin/store/writer/SaltStoreWriter.java
abuabraham-ttd: Added. And tested with previous implementation as well. It will only return newest snapshot in both instance

src/main/java/com/uid2/admin/store/writer/SaltStoreWriter.java
abuabraham-ttd: added the message in filter itself

src/test/java/com/uid2/admin/store/writer/EncryptedSaltStoreWriterTest.java
```diff
@@ -94,34 +96,92 @@ private RotatingSaltProvider.SaltSnapshot makeSnapshot(Instant effective, Instan
...
-        Integer idx = 0;
-        for (String line : contents.split("\n")) {
-            String[] entrySplit = line.split(",");
-            assertEquals(entries[idx].getId(), Long.parseLong(entrySplit[0]));
-            assertEquals(entries[idx].getSalt(), entrySplit[2]);
-            idx++;
-        }
+        String content = new String(encoded.readAllBytes(), StandardCharsets.UTF_8);
+        System.out.println(content);
+//        String contents = decryptInputStream(encoded, rotatingCloudEncryptionKeyProvider);
```
cody-constine-ttd, line 101: Can we remove commented out code?

cody-constine-ttd (approved):

pom.xml
atarassov-ttd: Suggest keeping unrelated changes to separate PRs.

PR #381, opened 2025-03-10
Reverting to old foreach logic

abuabraham-ttd:


src/main/java/com/uid2/admin/store/writer/SaltStoreWriter.java
```diff
@@ -48,23 +48,27 @@ protected List<RotatingSaltProvider.SaltSnapshot> getSnapshots(RotatingSaltProvi
             throw new IllegalStateException("Snapshots cannot be null");
```
abuabraham-ttd, line 48: making this behave same way as in [https://github.com/IABTechLab/uid2-admin/blob/345a708128c0b2937bc325b83082df5cd33f[…]/src/main/java/com/uid2/admin/store/writer/SaltStoreWriter.java](https://github.com/IABTechLab/uid2-admin/blob/345a708128c0b2937bc325b83082df5cd33fbb62/src/main/java/com/uid2/admin/store/writer/SaltStoreWriter.java)

PR #394, opened 2025-03-21
Using human readable dates

aulme:


src/main/java/com/uid2/admin/model/CloudEncryptionKeySummary.java
```diff
@@ -3,13 +3,24 @@
...
-        return new CloudEncryptionKeySummary(key.getId(), key.getSiteId(), key.getActivates(), key.getCreated());
+        return new CloudEncryptionKeySummary(
+                key.getId(),
+                key.getSiteId(),
+                toIsoTimestamp(key.getActivates()),
+                toIsoTimestamp(key.getCreated())
+        );
+    }
+
+    private static String toIsoTimestamp(Long epochSeconds) {
```
gmsdelmundo, line 23: Can we extract this to a repo-level util method? Might be useful for other pages as well in the future

aulme: Let's do that when we actually have the 2nd use case. YAGNI.

gmsdelmundo: But fair enough, we can just refactor it out once we start working on other stuff

gmsdelmundo (approved):

PR #399, opened 2025-03-25
Key rotation API with retention of 5 last expired keys

aulme:


src/main/java/com/uid2/admin/cloudEncryption/ExpiredKeyCountRetentionStrategy.java
```diff
@@ -0,0 +1,58 @@
...
+    private final Clock clock;
+    private final int expiredKeysToRetain;
+
+    public ExpiredKeyCountRetentionStrategy(Clock clock, int expiredKeysToRetain) {
+        this.clock = clock;
+        this.expiredKeysToRetain = expiredKeysToRetain;
+    }
+
+    @Override
+    public Set<CloudEncryptionKey> selectKeysToRetain(Set<CloudEncryptionKey> keysForSite) {
```
vishalegbert-ttd, line 24: This could be a lot simpler if we just sort by Created descending and take the top expiredKeysToRetain? 

I thought we didn't have a concept of "expired" keys?

vishalegbert-ttd: Perhaps expiredKeysToRetain should just be called keysToRetain

src/main/java/com/uid2/admin/vertx/service/CloudEncryptionKeyService.java
```diff
@@ -3,39 +3,85 @@
...
     }
 
     @Override
     public void setupRoutes(Router router) {
         router.get(Endpoints.CLOUD_ENCRYPTION_KEY_LIST.toString()).handler(
                 auth.handle(this::handleList, Role.MAINTAINER)
         );
+
+        router.post(Endpoints.CLOUD_ENCRYPTION_KEY_ROTATE.toString()).handler(
+                auth.handle(this::handleRotate, Role.MAINTAINER)
```
vishalegbert-ttd, line 53: Will this role allow for cron job to call it?

src/main/java/com/uid2/admin/cloudEncryption/ExpiredKeyCountRetentionStrategy.java
aulme: That is if we completely discount possibility of creating keys for the future. I guess we could do that with the new design. There is still code that creates some future keys on creation/update of operators in `CloudEncryptionKeyManager` - I'll remove it there as part of the next MR. Probably not a serious problem though - happy to change this part.

src/main/java/com/uid2/admin/vertx/service/CloudEncryptionKeyService.java
aulme: Ah good point. We actually don't have a role for cloud encryption stuff - encryption job is using PRIVATE_OPERATOR_SYNC. Both salt and encryption key rotation use `SECRET_ROTATION`, that may be reasonable to use here as well.

src/main/java/com/uid2/admin/cloudEncryption/ExpiredKeyCountRetentionStrategy.java
aulme: As a compromise and safeguard, I'll make this discard any keys which are in the future.

vishalegbert-ttd (approved):

PR #402, opened 2025-03-26
Remove code only used in its own test

aulme:


src/main/java/com/uid2/admin/Main.java
```diff
@@ -16,7 +16,7 @@
 import com.uid2.admin.managers.KeysetManager;
 import com.uid2.admin.cloudEncryption.CloudSecretGenerator;
 import com.uid2.admin.monitoring.DataStoreMetrics;
-import com.uid2.admin.managers.CloudEncryptionKeyManager;
+import com.uid2.admin.cloudEncryption.CloudEncryptionKeyManager;
```
gmsdelmundo, line 19: `cloudencryption`

gmsdelmundo (approved):

src/main/java/com/uid2/admin/Main.java
aulme: I'll rename in future commit. Don't really find that super readable but better stick to conventions.

PR #426, opened 2025-04-08
Creating cloud keys one hour in advance and keeping up to 10 existing…

aulme:
… keys

### DIFF
--- a/src/main/java/com/uid2/admin/cloudencryption/CloudEncryptionKeyManager.java
+++ b/src/main/java/com/uid2/admin/cloudencryption/CloudEncryptionKeyManager.java
@@ -36,7 +36,7 @@ public class CloudEncryptionKeyManager {
         this.planner = planner;
     }
 
-    // For any site that has an operator create a new key activating now
+    // For any site that has an operator create a new key activating in one hour
     // Keep up to 5 most recent old keys per site, delete the rest
     public void rotateKeys() throws Exception {
         boolean success = false;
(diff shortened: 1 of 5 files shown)


src/main/java/com/uid2/admin/cloudencryption/CloudKeyStatePlanner.java
```diff
@@ -17,6 +15,7 @@ public class CloudKeyStatePlanner {
         private final CloudSecretGenerator secretGenerator;
         private final Clock clock;
         private final CloudKeyRetentionStrategy keyRetentionStrategy;
+        private final int ONE_HOUR_IN_SECONDS = 60 * 60 * 1000;
```
CodyConstine, line 18: We should probably make this a const or a config value.

CodyConstine (approved):

src/main/java/com/uid2/admin/cloudencryption/CloudKeyStatePlanner.java
aulme: Will make it a const, don't think the number of seconds in an hour makes sense to be configurable :)

src/main/java/com/uid2/admin/cloudencryption/CloudEncryptionKeyManager.java
```diff
@@ -36,7 +36,7 @@ public CloudEncryptionKeyManager(
         this.planner = planner;
     }
 
-    // For any site that has an operator create a new key activating now
+    // For any site that has an operator create a new key activating in one hour
     // Keep up to 5 most recent old keys per site, delete the rest
```
gmsdelmundo, line 40: Should this now be 10 keys?

aulme: Good point, demonstrates how comments go out of date :)

gmsdelmundo (approved):

PR #428, opened 2025-04-09
Sample script to compare

abuabraham-ttd:


scripts/encryption_validator/readme.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Usage
+
+### Run the script:
+
+Login to AWS account
+`pip install requirements.txt` 
+`python script.py <key> <bucket> <region_name> [prefix]` 
+
+-   `key` – Required. Must start with `salt`. Example: `salts/salts.txt.1744152738842`
```
sunnywu, line 18: should key be renamed to object? key could be confused with UID2 (API/cloud encryption/etc) keys

sunnywu (approved):

scripts/encryption_validator/readme.md
sunnywu: @abuabraham-ttd still think you should rename key to something more specific

PR #430, opened 2025-04-09
Not deleting future keys on rotation since we now pre-create keys. Wr…

aulme:
…apped key rotation in a job.


src/main/java/com/uid2/admin/vertx/service/CloudEncryptionKeyService.java
```diff
@@ -38,8 +42,13 @@ public void setupRoutes(Router router) {
...
     private void handleRotate(RoutingContext rc) {
         try {
-            keyManager.rotateKeys();
-            rc.response().end();
+            jobDispatcher.enqueue(new CloudEncryptionKeyRotationJob(keyManager));
+            var isSuccess = jobDispatcher.executeNextJob().get();
+            if (isSuccess) {
+                rc.response().end();
+            } else {
+                rc.fail(500);
```
gmsdelmundo, line 50: Maybe we can add a message here about Key rotation failing?

gmsdelmundo (approved):

PR #435, opened 2025-04-10
Dont update version in encrypted data

abuabraham-ttd:


CodyConstine (approved):

pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.uid2</groupId>
```
abuabraham-ttd, line 7: DRAFT ; 

Added version everywhere. Want to see if it behaves as needed, will optimise and remove duplications

PR #456, opened 2025-05-08
Preliminary refactoring

aulme:


sophia-chen-ttd (approved):

src/main/java/com/uid2/admin/secret/SaltRotation.java
```diff
@@ -30,62 +35,94 @@ public Result rotateSalts(RotatingSaltProvider.SaltSnapshot lastSnapshot,
...
+                lastSnapshot.getFirstLevelSalt());
+        return Result.fromSnapshot(nextSnapshot);
+    }
+
+    private SaltEntry[] updateSalts(SaltEntry[] oldSalts, List<Integer> saltIndexesToRotate, long nextEffective) throws Exception {
+        var updatedSalts = new SaltEntry[oldSalts.length];
+
+        for (int i = 0; i < oldSalts.length; i++) {
+            var shouldRotate = saltIndexesToRotate.contains(i);
+            updatedSalts[i] = updateSalt(oldSalts[i], shouldRotate, nextEffective);
```
gmsdelmundo, line 58: Can we just skip the function call if `!shouldRotate` and set `updatedSalts[i] = oldSalts[i]`?

aulme: We can, but the whole point of this refactoring is to prepare for updating refreshFrom. We're planning to always calculate refreshFrom so we'd need to cycle through every salt.

sophia-chen-ttd: And we'll also need to update the previous from salt in here too

gmsdelmundo (approved):

PR #461, opened 2025-05-09
Recording refreshFrom, remove unnecessary interface

aulme:


src/test/java/com/uid2/admin/secret/SaltRotationTest.java
```diff
@@ -22,8 +23,18 @@ public class SaltRotationTest {
     @Mock private IKeyGenerator keyGenerator;
     private SaltRotation saltRotation;
 
-    private final LocalDate targetDate = LocalDate.of(2025, 1, 1);
-    private final Instant targetDateAsInstant = targetDate.atStartOfDay().toInstant(ZoneOffset.UTC);
+    private static final LocalDate targetDate = LocalDate.of(2025, 1, 1);
```
gmsdelmundo, line 26: By convention, `static final` should be all uppercase

i.e. `TARGET_DATE`, `ONE_DAY_EARLIER`, etc.

src/test/java/com/uid2/admin/secret/SaltRotationTest.java
```diff
@@ -57,25 +72,19 @@ private int countEntriesWithLastUpdated(SaltEntry[] entries, Instant lastUpdated
...
                 .withEntries(10, targetDateAsInstant)
-                .build(targetDateAsInstant,
-                        targetDateAsInstant.plus(7, ChronoUnit.DAYS));
+                .build(targetDateAsInstant, sevenDaysLater);
 
-        final ISaltRotation.Result result1 = saltRotation.rotateSalts(lastSnapshot, minAges, 0.2, targetDate);
+        var result1 = saltRotation.rotateSalts(lastSnapshot, minAges, 0.2, targetDate);
         assertFalse(result1.hasSnapshot());
-        final ISaltRotation.Result result2 = saltRotation.rotateSalts(lastSnapshot, minAges, 0.2, targetDate.minusDays(1));
+        var result2 = saltRotation.rotateSalts(lastSnapshot, minAges, 0.2, targetDate.minusDays(1));
```
gmsdelmundo, line 87: Same with some other parts, but you already have the `oneDayEarlier` variable set so you can just use that directly

Also wondering if we should just parametrize this test

src/test/java/com/uid2/admin/secret/SaltRotationTest.java
```diff
@@ -125,24 +131,22 @@ void rotateSaltsRotateSaltsFromOldestBucketOnly() throws Exception {
...
-                .withEntries(2, lastUpdated3)
-                .build(targetDateAsInstant.minus(1, ChronoUnit.DAYS),
-                        targetDateAsInstant.plus(6, ChronoUnit.DAYS));
-
-        final ISaltRotation.Result result = saltRotation.rotateSalts(lastSnapshot, minAges, 0.2, targetDate);
+        var lastSnapshot = SnapshotBuilder.start()
+                .withEntries(3, sixDaysEarlier)
+                .withEntries(5, fiveDaysEarlier)
+                .withEntries(2, fourDaysEarlier)
+                .build(targetDateAsInstant.minus(1, DAYS),
```
gmsdelmundo, line 138: Same here regarding the day variables you created previously

src/main/java/com/uid2/admin/secret/SaltRotation.java
```diff
@@ -63,19 +61,27 @@ private SaltEntry[] updateSalts(SaltEntry[] oldSalts, List<Integer> saltIndexesT
...
-                null,
+                refreshFrom,
                 null,
                 null,
                 null
         );
     }
 
+    private long calculateRefreshFrom(long lastUpdated, long nextEffective) {
+        long thirtyDaysMillis = 2592000000L;
```
sophia-chen-ttd, line 79: You can also use Duration.ofDays(30).toMillis();

aulme: This is done 1 million times in a loop so I wanted to reduce logic to minimum. Perhaps could move it to class level so it doesn't get recomputed every time.

src/test/java/com/uid2/admin/secret/SaltRotationTest.java
```diff
@@ -22,8 +23,18 @@ public class SaltRotationTest {
...
+
+    private static final Instant oneDayEarlier = targetDateAsInstant.minus(1, DAYS);
+    private static final Instant twoDaysEarlier = targetDateAsInstant.minus(2, DAYS);
+    private static final Instant threeDaysEarlier = targetDateAsInstant.minus(3, DAYS);
+    private static final Instant fourDaysEarlier = targetDateAsInstant.minus(4, DAYS);
+    private static final Instant fiveDaysEarlier = targetDateAsInstant.minus(5, DAYS);
+    private static final Instant sixDaysEarlier = targetDateAsInstant.minus(6, DAYS);
+    private static final Instant tenDaysEarlier = targetDateAsInstant.minus(10, DAYS);
+    private static final Instant sixDaysLater = targetDateAsInstant.plus(6, DAYS);
+    private static final Instant sevenDaysLater = targetDateAsInstant.plus(7, DAYS);
```
sophia-chen-ttd, line 37: would it be worth having a small function where we can put in the days to we want to add / subtract so this can be more flexible? When we start testing for rotations on intervals of 30s, these values will need to change anyway

src/test/java/com/uid2/admin/secret/SaltRotationTest.java
aulme: I'd rather make them non-static 😓

src/test/java/com/uid2/admin/secret/SaltRotationTest.java
aulme: The one day earlier variable is an Instant, while here we need a LocalDate, converting would be even longer.

Parametrization - probably will be riddled with if statements, maybe an exercise for another day.

src/test/java/com/uid2/admin/secret/SaltRotationTest.java
aulme: Missed this one, fixed

src/test/java/com/uid2/admin/secret/SaltRotationTest.java
aulme: Good point, we already have those.

src/test/java/com/uid2/admin/secret/SaltRotationTest.java
aulme: With @sophia-chen-ttd 's suggestion no longer need those.

sophia-chen-ttd (approved):

PR #473, opened 2025-05-13
sch-UID2-5352 honouring refresh from in salt rotation

sophia-chen-ttd:
Salts will only be considered for rotation if their refresh_from is equal to the target date. Feature is behind a feature switch, which can be enabled through the admin config file.

Unit test:
* salts that are old enough to rotate and are refreshing on target date will be considered for rotation
* salts that not old enough to rotate and are refreshing on the target date are not rotated
* salts that are old enough to rotate but are not refreshing on target date are not rotated
* refresh_from dates are updated appropriately

### DIFF
--- a/conf/local-config.json
+++ b/conf/local-config.json
@@ -12,7 +12,7 @@
   "keys_acl_metadata_path": "keys_acl/metadata.json",
   "salts_metadata_path": "salts/metadata.json",
   "salt_snapshot_location_prefix": "salts/salts.txt.",
-  "enable_salt_rotation_refresh_from": true,
+  "enable_salt_rotation_refresh_from": false,
   "operators_metadata_path": "operators/metadata.json",
   "enclaves_metadata_path": "enclaves/metadata.json",
   "partners_metadata_path": "partners/metadata.json",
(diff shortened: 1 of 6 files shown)


src/main/java/com/uid2/admin/secret/SaltRotation.java
```diff
@@ -17,11 +18,15 @@
 import static java.util.stream.Collectors.toList;
 
 public class SaltRotation {
+    private final static String ENABLE_REFRESH_FROM = "enable_salt_rotation_refresh_from";
```
gmsdelmundo, line 21: I think this goes into `AdminConst.java`

src/test/java/com/uid2/admin/secret/SaltRotationTest.java
```diff
@@ -34,11 +34,14 @@ private Instant daysLater(int days) {
         return targetDateAsInstant.plus(days, DAYS);
     }
 
-    @BeforeEach
```
gmsdelmundo, line 37: You could probably just keep the `@BeforeEach` and reconstruct `saltRotation` in the last function with `enableRefreshFrom` set to `true`

conf/local-config.json
```diff
@@ -12,6 +12,7 @@
   "keys_acl_metadata_path": "keys_acl/metadata.json",
   "salts_metadata_path": "salts/metadata.json",
   "salt_snapshot_location_prefix": "salts/salts.txt.",
+  "enable_salt_rotation_refresh_from": true,
```
gmsdelmundo, line 15: You can also add this in `default-config.json` and set to false

gmsdelmundo (approved):

aulme (approved):

PR #475, opened 2025-05-14
Logging salt age stats on rotation

aulme:


src/main/java/com/uid2/admin/secret/SaltRotation.java
```diff
@@ -94,86 +125,142 @@ private String calculatePreviousSalt(SaltEntry salt, boolean shouldRotate, long
...
-        Collections.shuffle(candidateIndexes);
-        candidateIndexes.stream().limit(maxIndexes).forEachOrdered(entryIndexes::add);
+        return candidateIndexes;
     }
 
-    private static boolean isBetween(long t, long minInclusive, long maxExclusive) {
-        return minInclusive <= t && t < maxExclusive;
+    private boolean isRotatable(long nextEffective, SaltEntry salt) {
+        if (this.isRefreshFromEnabled) {
+            if (salt.refreshFrom() == null) { // TODO: remove once refreshFrom is no longer optional
```
vishalegbert-ttd, line 204: Is it still optional? I thought we've populated this everywhere now?

src/main/java/com/uid2/admin/secret/SaltRotation.java
```diff
@@ -3,58 +3,89 @@
...
+        var nextSnapshot = new SaltSnapshot(
                 nextEffective,
                 nextExpires,
-                updatedSalts,
+                postRotationSalts,
                 lastSnapshot.getFirstLevelSalt());
         return Result.fromSnapshot(nextSnapshot);
     }
 
+    private List<Integer> findRotatableSaltIndexes(SaltEntry[] preRotationSalts, long nextEffective) {
```
vishalegbert-ttd, line 79: Is there a reason why we can't use salt bucket IDs rather than indexes?

src/main/java/com/uid2/admin/secret/SaltRotation.java
```diff
@@ -94,86 +125,142 @@ private String calculatePreviousSalt(SaltEntry salt, boolean shouldRotate, long
             return salt.currentSalt();
         }
         long age = nextEffective - salt.lastUpdated();
```
vishalegbert-ttd, line 127: Should we add a SaltEntry.ageInDays(asOf) method?

src/main/java/com/uid2/admin/secret/SaltRotation.java
```diff
@@ -94,86 +125,142 @@ private String calculatePreviousSalt(SaltEntry salt, boolean shouldRotate, long
...
+        var formattedDate = DateTimeFormatter.ofPattern("yyyy-MM-dd").format(targetDate);
+
+        var ages = new HashMap<Long, Long>(); // salt age to count
+        for (var salt : salts) {
+            long age = (nextEffective.toEpochMilli() - salt.lastUpdated()) / DAY_IN_MS;
+            ages.put(age, ages.getOrDefault(age, 0L) + 1);
+        }
+
+        for (var entry : ages.entrySet()) {
+            LOGGER.info("{} target-date={} age={} salts={}", logEvent, formattedDate, entry.getKey(), entry.getValue());
```
vishalegbert-ttd, line 223: perhaps:
salts={} target-date={} age={} num_salts={}

src/main/java/com/uid2/admin/secret/SaltRotation.java
```diff
@@ -94,86 +125,142 @@ private String calculatePreviousSalt(SaltEntry salt, boolean shouldRotate, long
...
+
+    private static SaltEntry[] onlySaltsAtIndexes(SaltEntry[] salts, List<Integer> saltIndexes) {
+        SaltEntry[] selected = new SaltEntry[saltIndexes.size()];
+        for (int i = 0; i < saltIndexes.size(); i++) {
+            selected[i] = salts[saltIndexes.get(i)];
+        }
+        return selected;
     }
 
+    private void logSaltCounts(LocalDate targetDate, Instant nextEffective, SaltEntry[] preRotationSalts, SaltEntry[] postRotationSalts, List<Integer> rotatableSaltIndexes, List<Integer> rotatedSaltIndexes) {
```
vishalegbert-ttd, line 235: Can we also track the number of salts that are refreshing at the target date? Will help us validate that on average 1/30 salts are up for refresh every day

src/main/java/com/uid2/admin/secret/SaltRotation.java
```diff
@@ -94,86 +125,142 @@ private String calculatePreviousSalt(SaltEntry salt, boolean shouldRotate, long
...
+        var rotatableSalts = onlySaltsAtIndexes(preRotationSalts, rotatableSaltIndexes);
```
vishalegbert-ttd, line 236: Can we pass in rotatableSalts, rotatedSalts (and saltsForRefresh) directly into this function?

src/main/java/com/uid2/admin/secret/SaltRotation.java
vishalegbert-ttd: could even just pass around arrays of salt entries and override its equal method if that works better

src/main/java/com/uid2/admin/secret/SaltRotation.java
aulme: Yeah, in practice we have that populated in every environment, there are still code paths for null case though. I'll remove it from here, we can clean up the others later.

src/main/java/com/uid2/admin/secret/SaltRotation.java
aulme: What's `salts` vs `num_salts` here? logEvent?

vishalegbert-ttd: salts would be the set of salts "rotatable-salts", "rotated-salts", etc. How were you planning on identifying the set of salts without a logfmt label, regex?

num_salts would be the number of salts

src/main/java/com/uid2/admin/secret/SaltRotation.java
aulme: We could, I wanted to be consistent with how `saltIndexesToRotate` is done, which is indexes. 

I guess that's done for performance reasons, so we can address specific indexes in the array of >1m rather than cycling through it. I can look at how much difference switching to objects across the board would make, might not be enough to matter.

src/main/java/com/uid2/admin/secret/SaltRotation.java
aulme: SaltEntry is in Shared and currently this isn't needed anywhere except Admin. I guess we might need this for Operator though so might be ok.

src/main/java/com/uid2/admin/secret/SaltRotation.java
aulme: Those are the rotatable salts

src/main/java/com/uid2/admin/secret/SaltRotation.java
aulme: RotatableSalts are saltsForRefresh

src/main/java/com/uid2/admin/secret/SaltRotation.java
vishalegbert-ttd: I think Operator checks salt age as well, so would be broadly useful

vishalegbert-ttd: We could just have one Admin helper function for this if we don't want to add it to shared

src/main/java/com/uid2/admin/salt/SaltRotation.java
```diff
@@ -0,0 +1,218 @@
...
+    ) throws Exception {
+        var preRotationSalts = lastSnapshot.getAllRotatingSalts();
+        var nextEffective = targetDate.asInstant();
+        var nextExpires = nextEffective.plus(7, ChronoUnit.DAYS);
+        if (nextEffective.equals(lastSnapshot.getEffective()) || nextEffective.isBefore(lastSnapshot.getEffective())) {
+            return Result.noSnapshot("cannot create a new salt snapshot with effective timestamp equal or prior to that of an existing snapshot");
+        }
+
+        // Salts that can be rotated based on their refreshFrom being at target date
+        var rotatableSalts = findRotatableSalts(preRotationSalts, targetDate);
```
sophia-chen-ttd, line 44: We can rename this to refreshableSalts so that it has more correlation to the refreshFrom field.

aulme: Renamed

src/main/java/com/uid2/admin/salt/SaltRotation.java
```diff
@@ -0,0 +1,218 @@
...
+    private long calculateRefreshFrom(SaltEntry salt, TargetDate targetDate) {
+        long multiplier = targetDate.saltAgeInDays(salt) / 30 + 1;
+        return salt.lastUpdated() + (multiplier * THIRTY_DAYS_IN_MS);
+    }
+
+    private String calculatePreviousSalt(SaltEntry salt, boolean shouldRotate, TargetDate targetDate) {
+        if (shouldRotate) {
+            return salt.currentSalt();
+        }
+        if (targetDate.saltAgeInDays(salt) < 90) {
```
vishalegbert-ttd, line 125: should this be less than or equal to? or less than?

aulme: I think this should be <, not <= since this includes the day on which we actually rotated the salt when the age would be 0. This results in people seeing previousSalt for 90 days exactly.

The current behaviour is also less than, we're even testing for it in L295.

src/test/java/com/uid2/admin/salt/helper/SaltBuilder.java
```diff
@@ -0,0 +1,62 @@
+package com.uid2.admin.salt.helper;
+
+import com.uid2.admin.salt.SaltRotation;
+import com.uid2.admin.salt.TargetDate;
+import com.uid2.shared.model.SaltEntry;
+
+import java.time.Instant;
+
+public class SaltBuilder {
```
gmsdelmundo, line 9: Do we want to use https://projectlombok.org/features/Builder instead?

src/test/java/com/uid2/admin/salt/helper/SaltBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+    private static int lastAutoId = 0;
+
+    private int id = lastAutoId++;
```
gmsdelmundo, line 12: Can we use atomics instead?

aulme: Sure

src/test/java/com/uid2/admin/salt/helper/SaltBuilder.java
aulme: Prefer not to:
1. This would leave builders in prod code which are only used in tests.
2. That's in Shared and we haven't actually done UID2-1750
3. They're not particularly flexible, if the builder can't hide uninteresting parts of building an object can as well just use new everywhere.

PR #498, opened 2025-06-03
No throw when backfillKeys() fails

clarkxuyang:
No throw when backfillKeys() fails

### DIFF
--- a/src/main/java/com/uid2/admin/cloudencryption/CloudEncryptionKeyManager.java
+++ b/src/main/java/com/uid2/admin/cloudencryption/CloudEncryptionKeyManager.java
@@ -72,7 +72,6 @@ public class CloudEncryptionKeyManager {
             LOGGER.info("Key backfill complete. Diff: {}", diff);
         } catch (Exception e) {
             LOGGER.error("Key backfill failed", e);
-            throw e;
         }
     }
 


src/main/java/com/uid2/admin/AdminConst.java
```diff
@@ -2,6 +2,7 @@
 
 public class AdminConst {
     public static String enableKeysetConfigProp = "enable_keysets";
+    public static final String throwOnS3WriteException = "throw_on_s3_write_exception";
```
aulme, line 5: This isn't really about S3 write exceptions, but more about exception during backfilling. We would still error on other S3 writes by admin.  

Also naming is inconsistent with existing conventions for naming constants (ALL_CAPS_WITH_UNDERSCORES).

src/main/java/com/uid2/admin/Main.java
```diff
@@ -291,8 +292,9 @@ public void run() {
                 }
             }
 
+            boolean bThrowOnS3WriteException = config.getBoolean(throwOnS3WriteException, true);
```
aulme, line 295: We don't use Hungarian notation in the current conventions. If the constant is renamed to all caps, we can just use the camel case version.

src/main/java/com/uid2/admin/vertx/service/OperatorKeyService.java
```diff
@@ -48,7 +49,9 @@ public class OperatorKeyService implements IService {
     private final IKeyGenerator keyGenerator;
     private final KeyHasher keyHasher;
     private final String operatorKeyPrefix;
+    
     private final CloudEncryptionKeyManager cloudEncryptionKeyManager;
+    private boolean bThrowOnS3WriteException;
```
aulme, line 54: Again, Hungarian notation.

src/main/java/com/uid2/admin/AdminConst.java
vishalegbert-ttd: I don't think we need a feature switch for this at all 🤷

PR #504, opened 2025-06-10
Add the ability to delete client-side keypairs

Ian-Nara:
UI and output look like:
keypairkeleteoutput

Asks for confirmation:
[
keypairdeleteconfirm
](url)

### DIFF
--- a/src/main/java/com/uid2/admin/vertx/service/ClientSideKeypairService.java
+++ b/src/main/java/com/uid2/admin/vertx/service/ClientSideKeypairService.java
@@ -76,6 +76,11 @@ public class ClientSideKeypairService implements IService, IKeypairManager {
                 this.handleUpdateKeypair(ctx);
             }
         }, new AuditParams(Collections.emptyList(), List.of("subscription_id", "name", "contact", "disabled")), Role.MAINTAINER, Role.SHARING_PORTAL));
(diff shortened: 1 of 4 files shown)


src/main/java/com/uid2/admin/vertx/service/ClientSideKeypairService.java
```diff
@@ -243,4 +266,27 @@ public ClientSideKeypair createAndSaveSiteKeypair(int siteId, String contact, bo
         return newKeypair;
 
     }
+
+    private JsonObject getRequestBody(RoutingContext rc) {
```
Ian-Nara, line 270: These two helper methods are to reduce the code duplication of retrieving a specific keypair, which is now done by both update and delete

src/test/java/com/uid2/admin/vertx/ClientSideKeypairServiceTest.java
```diff
@@ -556,6 +556,66 @@ void updateKeypairDisabledAndName(Vertx vertx, VertxTestContext testContext) thr
             testContext.completeNow();
         });
     }
+
+    @Test
```
mcollins-ttd, line 560: Can we add some tests where the user doesn't have the required roles?

PR #549, opened 2025-07-14
sch-UID2-5444 added feature switch for default min ages in salt rotation

sophia-chen-ttd:
* Added ENABLE_SALT_ROTATION_CUSTOM_AGE_THRESHOLDS feature switch to Admin, default to false
* Wrapped default age sequence (i.e. 90 days, 120 days, ... 390 days) behind feature switch

Tests:
* ENABLE_SALT_ROTATION_CUSTOM_AGE_THRESHOLDS == true means that we will use the custom min_ages given in the URL parameter
* ENABLE_SALT_ROTATION_CUSTOM_AGE_THRESHOLDS == false means that we use the default age sequence (i.e. 90 days, 120 days, ... 390 days)
* If ENABLE_SALT_ROTATION_CUSTOM_AGE_THRESHOLDS == true but there is no parameter defined in the URL, we return an error.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-admin</artifactId>
-    <version>6.9.0</version>
+    <version>6.9.1-alpha-202-SNAPSHOT</version>
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>


src/main/java/com/uid2/admin/vertx/service/SaltService.java
```diff
@@ -117,8 +116,15 @@ private void handleSaltRotate(RoutingContext rc) {
         try {
             final Optional<Double> fraction = RequestUtil.getDouble(rc, "fraction");
             if (fraction.isEmpty()) return;
-            final Duration[] minAges = RequestUtil.getDurations(rc, "min_ages_in_seconds");
-            if (minAges == null) return;
+
+            final Duration[] ageThresholds;
```
aulme, line 120: Let's log what thresholds we ended up using

src/main/java/com/uid2/admin/vertx/service/SaltService.java
```diff
@@ -36,6 +33,7 @@ public class SaltService implements IService {
     private final SaltStoreWriter storageManager;
     private final RotatingSaltProvider saltProvider;
     private final SaltRotation saltRotation;
+    private final Duration[] defaultSaltRotationAgeThresholds;
```
gmsdelmundo, line 36: You can directly just set this variable to be a `static final`

e.g.
```java
private static final Duration[] SALT_ROTATION_AGE_THRESHOLDS = generateThresholds(30, 390, 30);
```

or

```java
private static final Duration[] SALT_ROTATION_AGE_THRESHOLDS = new Duration[] { ... };
```

gmsdelmundo: If you do the former, remember to make `generateThresholds()` a static function

gmsdelmundo (approved):

PR #568, opened 2025-08-25
sch-UID2-5853 Added logs for key bucket count in salt rotation

sophia-chen-ttd:
After each salt rotation, logs:
* number of buckets that have rotated from salt to key
* total number of buckets that have a currentKey
* total number of buckets that have a currentSalt


src/main/java/com/uid2/admin/salt/SaltRotation.java
```diff
@@ -246,6 +247,22 @@ private void logSaltAges(String saltCountType, TargetDate targetDate, Collection
         }
     }
 
+    private void logBucketFormatCount(TargetDate targetDate, SaltEntry[] preRotationSalts, SaltEntry[] postRotationSalts) {
+        int newKeyBucketCounter = 0, totalKeyBucketCounter = 0, totalSaltBucketCounter = 0;
```
aulme, line 251: I think this deserves a comment - mention the migration, that the salts are the old format and we're transitioning to encryption keys. Mention that this is to monitor the migration.

src/main/java/com/uid2/admin/salt/SaltRotation.java
```diff
@@ -246,6 +247,22 @@ private void logSaltAges(String saltCountType, TargetDate targetDate, Collection
...
+
+        for (int i = 0; i < preRotationSalts.length && i < postRotationSalts.length; i++) {
+            var oldSalt = preRotationSalts[i];
+            var updatedSalt = postRotationSalts[i];
+
+            if (updatedSalt.currentKey() != null) totalKeyBucketCounter++;
```
aulme, line 257: Will we also do the previous salt/keys?

src/main/java/com/uid2/admin/salt/SaltRotation.java
```diff
@@ -246,6 +247,22 @@ private void logSaltAges(String saltCountType, TargetDate targetDate, Collection
...
+            if (updatedSalt.currentSalt() != null) totalSaltBucketCounter++;
+            if (updatedSalt.currentKey() != null && oldSalt.currentSalt() != null) newKeyBucketCounter++;
+        }
+
+        LOGGER.info("Salt rotation bucket format: target_date={} new_key_bucket_count={} total_key_bucket_count={} total_salt_bucket_count={}",
```
aulme, line 262: Could be clearer that those are formats of buckets rotated just now, not for all bucket.

sophia-chen-ttd: These are for all the buckets in snapshot, including those that haven't rotated, whereas the new_key_bucket_count are for buckets that have rotated just now

src/main/java/com/uid2/admin/salt/SaltRotation.java
sophia-chen-ttd: I think we don't need to monitor those - I'm not sure how helpful they would be in monitoring the migration?

aulme: We should start seeing the first "previous key" ~1 year after we start migration and we should stop seeing "previous salt" soon after that, at least by day 420.

PR #615, opened 2026-04-01
UID2-1750: add explicit Lombok dependency (was relying on transitive)

sunnywu:
## Summary
- Add Lombok 1.18.34 as `provided` + `optional` dependency explicitly in `pom.xml`
- Register Lombok as an explicit `annotationProcessorPath` in `maven-compiler-plugin`

## Why
uid2-admin uses Lombok annotations (`@Getter`, `@AllArgsConstructor`, `lombok.val` in 9 files) but was silently relying on Lombok leaking transitively from uid2-shared's incorrect `compile`-scope declaration. The companion PR IABTechLab/uid2-shared#610 fixes uid2-shared to declare Lombok as `provided`+`optional`, stopping the transitive leak. uid2-admin must therefore declare Lombok explicitly. See [UID2-1750](https://thetradedesk.atlassian.net/browse/UID2-1750).

## Test plan
- [x] `mvn compile` passes (151 source files, 0 errors)
- [x] `mvn test` passes (698 tests, 0 failures, 1 skipped)
- [x] `mvn dependency:tree | grep lombok` shows `provided` scope (not `compile`)

> Merge uid2-shared companion PR first, then this one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,7 +52,6 @@
             <artifactId>lombok</artifactId>
             <version>1.18.34</version>
             <scope>provided</scope>
-            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>io.vertx</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -48,6 +48,13 @@
             <version>${uid2-shared.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.34</version>
+            <scope>provided</scope>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
             <groupId>io.vertx</groupId>
             <artifactId>vertx-core</artifactId>
             <version>${vertx.version}</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -60,6 +47,13 @@
             <artifactId>uid2-shared</artifactId>
             <version>${uid2-shared.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.34</version>
+            <scope>provided</scope>
```
jon8787, line 67: I think we don't need `provided` and `optional` here because uid2-admin is a leaf project

sunnywu: claude said:

 - optional — agree. optional prevents the dependency from being inherited by downstream projects. Since uid2-admin is a leaf application (nothing    
  depends on it as a library), it has no downstream consumers, so optional is indeed pointless here.
  - provided — disagree. provided is actually correct for lombok regardless of whether it's a leaf project. Lombok is a compile-time annotation        
  processor — it generates bytecode at compile time and is not needed at runtime. Using provided keeps it out of the final fat JAR, which is the right 
  behavior. This is standard practice for lombok in Maven projects.                                                                                    
                                                                                                                                                       
  The reviewer's "leaf project" reasoning only fully applies to optional. For provided, the reasoning is about compile-time vs runtime, not about      
  dependency inheritance.

